PR #41, opened 2018-06-24
#39 error Handling

tko22:
Resolves #39

PR #44, opened 2018-06-24
Can filter by status and assign status to Candidates

### LINKED ISSUE #23: Filter by Status
Make the filter by status functional in the frontend. User should be able to filter candidates by the following categories:
* Accepted
* Pending
* Interviewing
* Rejected
* Everyone

shreyshrey1:
Resolves #23 

Bugs:
- Cards don't render correctly when under 3 candidates displayed
- Initializing statuses to "pending" endpoint doesn't work, commented in code. I had to go through and manually assign each one to pending.

### DIFF
--- a/backend/src/index.js
+++ b/backend/src/index.js
@@ -1,6 +1,7 @@
 const express = require('express')
 const bodyParser = require('body-parser')
 const mongoose = require('mongoose')
+const querystring = require('querystring')
 const { Candidate, Stats, Match } = require('./models')
 const { errorWrap, getStats } = require('./utils')
 const cors = require('cors')
@@ -23,12 +24,17 @@ app.get('/', (req, res) => {
 app.get(
   '/candidates',
   errorWrap(async (req, res) => {
-    try {
-      const candidates = await Candidate.find()
-      res.json({ result: candidates })
-    } catch (err) {
-      res.json(400, { message: err.message })
+    let candidates
+    if (req.query.status) {
+      if (req.query.status == 'everyone') {
+        candidates = await Candidate.find()
+      } else {
+        candidates = await Candidate.find({ status: req.query.status })
+      }
+    } else {
+      candidates = await Candidate.find()
     }
+    res.json({ result: candidates })
   })
 )
 
@@ -112,7 +118,7 @@ app.post(
 app.get(
   '/parse',
   errorWrap(async (req, res) => {
-    const wb = XLSX.readFile('candidates.xlsx')
+    const wb = XLSX.readFile(__dirname + '/candidates.xlsx')
     const ws = wb.Sheets[wb.SheetNames[0]]
     var i = 0
     for (var elm in ws) {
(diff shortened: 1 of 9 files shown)


backend/src/index.js
```diff
@@ -125,6 +125,65 @@ app.get(
   })
 )
 
+app.get(
+  '/set-status/:status/:id',
+  errorWrap(async (req, res) => {
```
tko22: as a rule of thumb, any endpoint that modifies the database should be a POST or PUT request. Since this sets the status of a certain candidate, let's have the URL be something like `/candidates/:id/status` and make it a POST request

backend/src/index.js
```diff
@@ -125,6 +125,65 @@ app.get(
...
+        case "rejected":
+          await Candidate.findByIdAndUpdate(req.params.id, {status: "rejected"});
+          break;
+        default:
+          response = "Invalid status, please try again"
+      }
+      res.json({result: response})
+    } catch (err) {
+      res.status(400).json({ message: err.message })
+    }
```
tko22: Sorry for providing a bad example previously, but `errorWrap` catches any errors within the async function. This is what #41 is doing. errorWrap would catch the error and the errorHandler will handle, thus, removing the need for try...catch blocks. Let's remove the try...catch blocks for all of your new and modified endpoints. #41 removes try...catch blocks for the previously written endpoints.

backend/src/index.js
```diff
@@ -125,6 +125,65 @@ app.get(
...
+// Following function doesn't work, not sure why...
+app.get(
+  '/initialize-status',
+  errorWrap(async (req, res) => {
+    try {
+      Candidate.update({}, { $set: { status: 'pending' }})
+      res.send("Done updating")
+    } catch (err) {
+      res.status(400).json({ message: err.message })
+    }
```
tko22: same as above. Also if it doesn't work, let's delete it :)

backend/src/index.js
```diff
@@ -125,6 +125,65 @@ app.get(
...
+  })
+)
+
+app.get(
+  '/filter-status/:status',
```
tko22: this pulls from REST resource naming. Since you are returning certain candidates based off of their status. The URL endpoint should be part of /candidates, where status is one of the query parameters (stuff that's after `?` ) It would be something like `/candidates?status=pending` 

http://www.restapitutorial.com/lessons/restquicktips.html are some quick tips on dsigning a RESTful api

frontend/src/components/candidateCard.js
```diff
@@ -14,48 +16,78 @@ type Props = {
...
-          <span className="highlight">{candidate.role}</span>
-        </p>
-        <p>
-          Hours: <span className="highlight">{candidate.timeCanDevote}</span>
-        </p>
-      </div>
-    </CardBody>
-  </Card>
-)
+class CandidateCardComponent extends Component {
```
tko22: Let's add the flow <Props> to Component

tko22: Look at some of the other components to see how that works

tko22:
Filtering in client side could also be another possibility. Up to you though. or you could do both

backend/src/index.js
```diff
@@ -23,12 +24,17 @@ app.get('/', (req, res) => {
...
   '/candidates',
   errorWrap(async (req, res) => {
-    try {
-      const candidates = await Candidate.find()
-      res.json({ result: candidates })
-    } catch (err) {
-      res.json(400, { message: err.message })
+    let candidates
+    if (req.query.status) {
+      if (req.query.status == 'everyone') {
```
tko22: this is a little redundant since it returns the same thing if you don't include `?status`. but if you have a use case for it, that's fine. Just watch out for this stuff

tko22: and resolve your merge conflicts

PR #53, opened 2018-06-27
Move candidate endpoints and face mash endpoints to their own separate files

### LINKED ISSUE #45: logically Separate Endpoints in different files with router in express.js
use express.Router https://expressjs.com/en/guide/routing.html

candidates endpoints should be in one file, face mash should be in another. define those endpoints using express.Router

angadgarg25:
Resolves #45 

Endpoints are now split into different files for better design and organization. Original functionality maintained. (My git got a little messed up so a bunch of old commits that are already merged are showing up).


backend/package.json
```diff
@@ -14,6 +14,7 @@
   "dependencies": {
     "body-parser": "^1.18.2",
     "cors": "^2.8.4",
+    "dotenv": "^6.0.0",
```
tko22: I don't think this should be here. It should be installed globally. If you ran into issues using dotenv, just delete this line and you should still have it installed.

PR #54, opened 2018-06-27
Grouped endpoints and separated into different files

### LINKED ISSUE #45: logically Separate Endpoints in different files with router in express.js

angadgarg25:
Resolves #45 

Endpoints are now split into different files for better design and organization. Original functionality maintained.

### DIFF
--- a/backend/src/candidates.js
+++ b/backend/src/candidates.js
@@ -1,0 +1,83 @@
+const express = require('express')
+const router = express.Router()
+const { errorWrap } = require('./middleware')
+const { Candidate } = require('./models')
+
+router.get(
+  '/',
+  errorWrap(async (req, res) => {
+    let candidates
+    if (req.query.status) {
+      if (req.query.status == 'everyone') {
+        candidates = await Candidate.find()
+      } else {
+        candidates = await Candidate.find({ status: req.query.status })
+      }
+    } else {
+      candidates = await Candidate.find()
+    }
+    res.json({ result: candidates })
+  })
+)
+
+router.post(
+  '/',
+  errorWrap(async (req, res) => {
+    const c = new Candidate({
+      name: 'Tim',
+      email: 'other@gmail.com',
+      graduationDate: '2018',
+      major: 'CompE',
+      resumeID: 'resume2.pdf',
+      role: 'SWE'
+    })
+    await c.save()
+  })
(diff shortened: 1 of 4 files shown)


backend/src/candidates.js
```diff
@@ -0,0 +1,46 @@
+const express = require('express')
+const candidates = express.Router()
```
tko22: i think this should be named router, just to make it more explicit.

backend/src/candidates.js
```diff
@@ -0,0 +1,46 @@
...
+
+candidates.get(
+  '/:candidateId',
+  errorWrap(async (req, res) => {
+    const candidate = await Candidate.findById(req.params.candidateId)
+    res.json({ result: candidate })
+  })
+)
+
+module.exports = candidates
```
tko22: even if you do `module.exports = router` you can still do a `const candidates = require('./candidates')`. the names don't have to match up!

backend/src/matchCandidates.js
```diff
@@ -0,0 +1,44 @@
+const express = require('express')
+const matchCandidates = express.Router()
```
tko22: same as above

backend/src/index.js
```diff
@@ -22,8 +22,8 @@ app.get('/', (req, res) => {
   res.send('hi')
 })
 
-app.use('/candidates', candidates)
-app.use('/matchCandidates', matchCandidates)
+app.use('/', candidates)
+app.use('/', matchCandidates)
```
angadgarg25: @tko22 This feels a little bad but since the endpoints in each file don't share a common prefix anymore, I thought this was the best way. The other alternative I can think of is renaming the paths of related endpoints to share a common prefix.

tko22: ah sorry for the confusion. I didn't mean changing it in `index.js`. just renaming the router to be `router = express.Router()` in both `candidates.js` and `matchCandidates.js` instead of it being `candidates = express.Router()`. I agree with you that it should be 
```Javascript
app.use('/candidates', candidates)
app.use('/matchCandidates', matchCandidates)
```

PR #64, opened 2018-07-12
Persist Facemash Candidate Pairings

### LINKED ISSUE #11: Refreshing Facemash shouldn't change candidates shown
the data should persist using Redux and [Redux Persist](https://www.npmjs.com/package/redux-persist) middleware

### LINKED ISSUE #12: Picking a candidate should change generate new candidates
Can't just refresh the page since it will be fixed in #11

angadgarg25:
Resolves #11, Resolves #12 

Maintained candidate pairing stored in redux through LocalStorage API. Candidate pairings will persist even if the app is shutdown and restarted. To delete save data in Chrome go here: chrome://settings/siteData and search for localhost and delete the LocalStorage data.


tko22:
Let's update to Next6 but also this (https://egghead.io/lessons/javascript-redux-persisting-the-state-to-the-local-storage) might be worth a look. It's from the creator of redux and doesn't require any external dependencies. If that suits your needs, then I'd get rid of ReduxPersist. As a rule of thumb, you should only add dependencies if you really need them. 

for the upgrade to Nextjs 6, make it a separate PR and make sure the `yarn.lock` file gets updated as well.

tko22:
i do like how you made a PR for this and asked everyone for their opinions. Great Job :)

frontend/src/pages/facemash.js
```diff
@@ -34,6 +34,10 @@ class FaceMash extends Component<Props> {
       error: false,
       message: ''
     }
+    const { candidates, generateMatchData } = this.props
+    if (candidates == null || candidates.length != 2) {
+      generateMatchData(this.state.results.candidate1, this.state.results.candidate2)
+    }
```
angadgarg25: Since the results come from `getInitialProps` which is called from the server, a candidate pairing request will happen every time regardless of whether the client already has a saved pairing locally, because the server has no way of knowing whether the client has one.

tko22: yes, the server generate matches and validate whether a match has been created when a client submits a match pick. Can we move it away from `getIntialProps`?

frontend/src/pages/facemash.js
```diff
@@ -76,29 +88,30 @@ class FaceMash extends Component<Props> {
     if (this.props.error) {
       return <div>Bad Fetch. Try again</div>
     }
-    return (
+    const { candidates } = this.props
+    return candidates && candidates.length == 2 ? (
```
tko22: nice job checking.

frontend/src/pages/index.js
```diff
@@ -22,8 +25,23 @@ type Props = {
   result: {}
 }
 
+function mapStateToProps(state) {
+  return {
+    candidates: state.candidates
+  }
+}
```
tko22: some cool syntax you could is:

const mapStateToProps =(state) => (
    {candidates: state.candidates}
)

i think

tko22: you don’t have to change it to that though

frontend/src/store/localStorage.js
```diff
@@ -0,0 +1,25 @@
...
+export const loadState = () => {
+  try {
+    const serializedState = localStorage.getItem(STATE_STORAGE)
+    if (serializedState === undefined || serializedState === null) {
+      localStorage.setItem(STATE_STORAGE, initialState)
+      return initialState
+    }
+    return JSON.parse(serializedState)
+  } catch (err) {
+    return initialState
```
tko22: log an error here?

frontend/src/store/localStorage.js
```diff
@@ -0,0 +1,25 @@
...
+  }
+}
+
+export const saveState = state => {
+  try {
+    const serializedState = JSON.stringify(state)
+    localStorage.setItem(STATE_STORAGE, serializedState)
+  } catch (err) {
+    // fill
```
tko22: yes

frontend/src/pages/facemash.js
```diff
@@ -30,75 +29,78 @@ class FaceMash extends Component<Props> {
...
-        return { error: 'Bad Request' }
-      }
-      return { result }
-    } catch (err) {
-      console.log('Candidate Page error: ', err.message)
-      return { error: 'Bad Request' }
+  async getNewMatch() {
+    const { result } = await getCandidateMatch()
+    if (result === undefined) {
+      console.log('Could not get new FaceMash match')
```
tko22: if `result` is undefined, the function still continues after logging that, meaning it will fail in line 44. maybe alert the user that it wasn't able to get a match by rendering a placeholder saying `Error: Couldn't get new FaceMash match. Please refresh` or using a popup/alert/toast?

frontend/src/pages/facemash.js
```diff
@@ -30,75 +29,78 @@ class FaceMash extends Component<Props> {
...
       const res = await setMatchWinner(
-        candidate1._id,
-        candidate2._id,
-        winner === '0' ? candidate1._id : candidate2._id,
+        candidates[0]._id,
+        candidates[1]._id,
+        winner === '0' ? candidates[0]._id : candidates[1]._id,
         matchID
       )
+      console.log('madeuit')
```
tko22: remove this, unless it's necessary

frontend/src/pages/facemash.js
```diff
@@ -30,75 +29,78 @@ class FaceMash extends Component<Props> {
...
+              <Candidate candidate={candidates[1]} />
               <button name="1" className="btn btn-info" onClick={this.handleClick}>
                 Pick
               </button>
             </div>
           </div>
         </Container>
       </div>
-    )
+    ) : null
```
tko22: instead of returning null. return a place holder stating the error. like this https://github.com/hack4impact-uiuc/love-without-boundaries/blob/3c8ceb3d76c44ce15b8257cc147ae6cf3b092476/react_frontend/src/pages/StudentPage.js#L60

which would render something like this 
image

frontend/src/reducers/initialState.js
```diff
@@ -1,5 +1,6 @@
 const initialState = {
-  candidates: ['abc']
+  candidates: ['abc'],
```
tko22: why is matchID `null` but candidates has `abc`?

frontend/src/pages/facemash.js
```diff
@@ -11,11 +11,10 @@ import Candidate from '../components/candidateBox'
 import { Container } from 'reactstrap'
 type Props = {}
```
tko22: let's define the types of the props here. https://flow.org/en/docs/types/

doesn't have to be too specific, can just be type Object or something like that. This will improve readability, especially when we come back to this in later semesters

frontend/src/pages/facemash.js
```diff
@@ -30,74 +33,79 @@ class FaceMash extends Component<Props> {
...
+        candidates[0]._id,
+        candidates[1]._id,
+        winner === '0' ? candidates[0]._id : candidates[1]._id,
         matchID
       )
       if (res.success) {
         this.setState({
           success: 'Successfully Submitted'
         })
       }
```
tko22: I know you didn't change the backend endpoint for this but:
- the backend endpoint shouldn't return `{ success: true }`, instead it should return something like this: 
``` Javascript
{
  "code": 200, 
  "message": "", 
  "result": {}, 
  "success": true
}
```
I'd also add an else statement if the endpoint returns success `false`, probably just `console.log`and make an alert pop up.

frontend/src/pages/facemash.js
```diff
@@ -30,74 +33,79 @@ class FaceMash extends Component<Props> {
...
+    if (candidates == null || candidates.length != 2) {
+      this.getNewMatch()
     }
   }
 
   handleClick = async e => {
     const winner = e.target.name
     try {
-      const { candidate1, candidate2, matchID } = this.state.results
+      const { candidates, matchID } = this.props
```
tko22: I would also check whether `candidates` and `matchID` are null here and not send the POST request to backend.

frontend/src/pages/facemash.js
```diff
@@ -30,74 +33,85 @@ class FaceMash extends Component<Props> {
...
+        candidates[1]._id,
+        winner === '0' ? candidates[0]._id : candidates[1]._id,
         matchID
       )
       if (res.success) {
         this.setState({
           success: 'Successfully Submitted'
         })
+      } else {
+        console.error('Match not successfully submitted')
```
tko22: add a ‘window.alert(“text here”)’ 

https://www.w3schools.com/jsref/met_win_alert.asp

PR #66, opened 2018-07-16
Filter candidates and redux

shreyshrey1:
Resolves a bunch of issues
Still pretty buggy, work in progress.


backend/src/api/candidates.js
```diff
@@ -38,10 +38,40 @@ router.post(
...
-  '/:candidateId',
+  '/initialize-status',
   errorWrap(async (req, res) => {
-    const candidate = await Candidate.findById(req.params.candidateId)
-    res.json({ result: candidate })
+    await Candidate.update({}, { status: 'pending' }, { multi: true }, err => {
+      if (err) {
+        console.log(err.message)
+      }
+    })
```
tko22: `errorWrap` catches all of the errors, and `middleware/errorHandler` handles it so you dont need this part

backend/src/api/candidates.js
```diff
@@ -38,10 +38,40 @@ router.post(
...
+    res.send('Set all status to pending')
+  })
+)
+
+router.get(
+  '/initialize-year',
+  errorWrap(async (req, res) => {
+    const candidates = await Candidate.find()
+    const years = ['freshman', 'sophomore', 'junior', 'senior']
```
tko22: the application form has years by graduation semester eg: `Spring 2019`, `Fall 2018`. Let's use that instead

shreyshrey1: But those constantly change

backend/src/api/candidates.js
```diff
@@ -38,10 +38,40 @@ router.post(
...
+    })
+    res.send('Initialize Years')
+  })
+)
+
+router.get(
+  '/initialize-role',
+  errorWrap(async (req, res) => {
+    const candidates = await Candidate.find()
+    const roles = ['software engineer', 'product manager', 'tech lead', 'community director']
```
tko22: Let's match these roles to the ones in the application Form. The first letter should capitalized, I believe.

yarn-error.log
```diff
@@ -0,0 +1,41 @@
+Arguments:
```
tko22: remove this file. It shouldn't be committed and add it to `.gitignore`

frontend/src/utils/enums.js
```diff
@@ -0,0 +1,20 @@
...
+  DENIED: 'rejected',
+  INTERVIEWING: 'interviewing',
+  PENDING: 'pending'
+}
+
+export const rolesenum = {
+  SWE: 'software engineer',
+  PM: 'product manager',
+  TL: 'tech lead',
+  CD: 'community director'
```
tko22: same comment as above, where you should change these enums to match what's on the Application form.

frontend/src/reducers/initialState.js
```diff
@@ -1,5 +1,30 @@
...
-  candidates: ['abc']
+  facemash: {
+    candidates: 'abc'
+  },
+  candidateListPage: {
+    candidates: [],
+    candidatesLoading: false,
+    candidatesError: null,
+    filters: {
+      years: [yearsenum.FRESHMAN, yearsenum.SOPHOMORE, yearsenum.JUNIOR, yearsenum.SENIOR],
```
tko22: remember to change this once the enums change.

frontend/src/components/filterComponent.js
```diff
@@ -0,0 +1,131 @@
...
+    })
+  }
+  render() {
+    const years = [yearsenum.FRESHMAN, yearsenum.SOPHOMORE, yearsenum.JUNIOR, yearsenum.SENIOR]
+    const roles = [rolesenum.SWE, rolesenum.TL, rolesenum.CD, rolesenum.PM]
+    const statuses = [
+      statusenum.PENDING,
+      statusenum.ACCEPTED,
+      statusenum.DENIED,
+      statusenum.INTERVIEWING
```
tko22: I like the enums :)

frontend/src/actions/actionCreators.js
```diff
@@ -13,3 +40,15 @@ function newMatch(id_1, id_2) {
 export function generateMatchData(id_1, id_2) {
   return dispatch => dispatch(newMatch(id_1, id_2))
 }
+
+export const fetchCandidates = () => {
+  return dispatch => {
+    dispatch(fetchCandidatesBegin())
+    return getAllCandidates()
```
tko22: i wonder if we could async await with this? idk if it's worth a shot though

frontend/src/pages/index.js
```diff
@@ -17,59 +17,69 @@ import {
...
+import { yearsenum, statusenum, rolesenum } from '../utils/enums'
 
 type Props = {
-  result: {}
+  candidates: Array<any>,
+  loading: boolean,
+  error: boolean,
+  filters: Object,
+  sort: Object
+}
```
tko22: yay! Types!

frontend/src/pages/index.js
```diff
@@ -17,59 +17,69 @@ import {
...
+
+const mapStateToProps = state => {
+  return {
+    candidates: state.candidateListPage.candidates,
+    loading: state.candidateListPage.candidatesLoading,
+    error: state.candidateListPage.candidatesError,
+    filters: state.candidateListPage.filters,
+    sort: state.candidateListPage.sort
+  }
```
tko22: cool thing you could do is wrap 45-51 in `( )` instead of `{ }` and you can remove the return statement.

so..
```Javascript
const asdfasdf = state => ({text:"asdf"})
```
up to you though

frontend/src/pages/index.js
```diff
@@ -17,59 +17,69 @@ import {
...
+  componentDidMount() {
+    this.props.fetchCandidates()
   }
 
   render() {
-    if (this.props.error) {
+    let { candidates, error, loading, filters, sort } = this.props
+    const statusFilter = this.props.filters.statuses
+    const roleFilter = this.props.filters.roles
+    const yearFilter = this.props.filters.years
```
tko22: you deconstructed props 71, lets use those variables then! so instead of `this.props.filters.years` you could do `filters.years`

frontend/src/pages/index.js
```diff
@@ -17,59 +17,69 @@ import {
...
+    candidates = candidates.filter(candidate => {
+      return (
+        statusFilter.includes(candidate.status) &&
+        roleFilter.includes(candidate.role) &&
+        yearFilter.includes(candidate.year)
+      )
+    })
+    if (error) {
       return <div>Bad Fetch. Try again</div>
     }
```
tko22: this should be handled right after you get `error` instead of processing some other stuff.

frontend/src/store/appStore.js
```diff
@@ -1,13 +1,10 @@
...
-
+import { composeWithDevTools } from 'redux-devtools-extension'
 export default function configureStore() {
   return createStore(
     recruitmentApp,
-    typeof window !== 'undefined' &&
-      window.__REDUX_DEVTOOLS_EXTENSION__ &&
-      window.__REDUX_DEVTOOLS_EXTENSION__(),
-    applyMiddleware(thunk)
+    /* preloadedState, */ composeWithDevTools(applyMiddleware(thunk))
```
tko22: this file may have some merge conflicts with angad's pr... we'll see. nothing wrong about this though. just thought you would like to know that.

shreyshrey1:
Should work now, feel free to pull and test more extensively

frontend/src/store/appStore.js
angadgarg25: ^ yeah i had to do this composition thing too. I think your stuff should work with my version of this file since I did this and added other things so merging shouldn't be too difficult.

shreyshrey1: Yeah the reason I changed it was before the preloadedState was being set to false so that was messing up things when the site first loaded.

Issue #20, opened 2018-06-17
Candidates Page & keys


tko22:
includes #9, #10, #17 
- adding interviews form will include a lot of things, look at the rubric 
- build key pages and backend verification
- work with candidates to generate categories
- split off and work with candidates list page team + facemash team
- easily show the state of a candidate (whether they are declined/accepted/pending/etc)
Remember that this page should look different before interviews, after interviews, during/after deliberations and decisions.
- show data from facemash

HRimaw2:
Candidates Page & keys #20 Breakdown

https://docs.google.com/document/d/1lnRazAewcm25Gf6idaGMm-AHDPMPbLbtlvuh41ln1TQ/edit

tko22:
I love it :)

tko22:
for the interview form, something you could do is add Form Validation if you have time. https://reactstrap.github.io/components/form/#app Scroll down a bit

PR #75, opened 2018-08-01
Now Deployment

### LINKED ISSUE #74: Deployment
🚀 Deploy everything

CodeBrew28:
Resolves #74

The frontend is now deployed on https://hack4impact-recruitment-frontend.now.sh/ and the backend on https://hack4impact-recruitment-backend.now.sh/. It runs differently in development mode, so the dev experience doesn't change.


backend/.dockerignore
```diff
@@ -0,0 +1,5 @@
+*
```
tko22: nice!

backend/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM mhart/alpine-node:10.7.0 as test
+WORKDIR /usr/src/app
+COPY package.json .
+COPY yarn.lock .
+RUN yarn install
+COPY . .
```
tko22: I don't think you'll need `COPY package.json` or `COPY yarn.lock` because it's copied here.

backend/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+RUN yarn test
```
tko22: does it fail the build if the tests fail? We also already have our tests run in the ci

backend/now.json
```diff
@@ -0,0 +1,10 @@
+{
+    "type": "docker",
+    "alias": "h4i-recruitment-backend.now.sh",
+    "env": {
+        "MONGO_URL": "@h4i-mongo-uri"
```
tko22: is this stored in the hack4impact now team?

backend/src/index.js
```diff
@@ -2,6 +2,7 @@ const app = require('./app')
 const mongoose = require('mongoose')
 
 // connect to mongoose
+console.log(process.env)
```
tko22: let's remove this

frontend/now.json
```diff
@@ -0,0 +1,7 @@
+{
+    "type": "npm",
```
tko22: is there a yarn? because npm wouldn't look at yarn.lock

backend/Dockerfile
angadgarg25: As far as I know, if we removed those two lines and put the `yarn install` after the `COPY . .` then every time we changed the source code and rebuilt the container it would re-run yarn install. If we leave this as is, it will only re-run `yarn install` when we re-build the container if the `package.json` or `yarn.lock` is edited.

tko22:
yep megha and hana are working on auth in #20

backend/Dockerfile
CodeBrew28: yep, and it should fail if the tests are not passing as we don't want to deploy it in that case.

frontend/now.json
CodeBrew28: Now supports 3 types of deployments: static, npm, and docker. npm vs yarn is just a package manager difference. This does mean though that internally it calls "npm install" and generates a "package-lock.json" file. The other solution will be making it a docker deployment. I can look into doing this, and possibly make another PR.

PR #78, opened 2018-08-11
add navbar + a little cleanup

tko22:
no issue

image


tko22:
Yeah, just changed it. I added it to split the imports from external packages and from our own code. But it doesn't look that great in javascript.

Issue #69, opened 2018-07-18
Standardize Backend Endpoint Responses


tko22:
https://hack4impact.gitbook.io/wiki/untitled/our-rest-api-specification
```Javascript
{
  "code": 200, 
  "message": "", 
  "result": {}, 
  "success": true
}
```

tko22:
we could make an express middleware for this instead.

tko22:
we do the same thing with flask-boilerplate with `create_response` https://github.com/tko22/flask-boilerplate/blob/88b9077dad74e7dd30404adffb2bf3ec2cc0bb2e/api/core.py#L27-L50

PR #83, opened 2018-08-12
Interview Backend

HRimaw2:
Tested backend in Postman, documentation is available in backend/docs/interview_schema.md
I'd like reviewing and comments, I haven't had much backend experience before


backend/src/api/interview.js
```diff
@@ -0,0 +1,129 @@
...
+router.get(
+  '/:interview_id',
+  errorWrap(async (req, res) => {
+    const retInterview = await Interview.findById(req.params.interview_id)
+    res.json({ interview: retInterview })
+  })
+)
+
+router.post(
+  '/add-interview',
```
tko22: this can just be a post request with `/interview/`. so it'll look like `router.post('/'` here

backend/src/api/interview.js
```diff
@@ -0,0 +1,129 @@
...
+        interviewer_key: interviewerKey,
+        overall_score: score,
+        candidate_id: candidateId,
+        general_notes: genNotes,
+        sections: reqSections
+      })
+      await interview.save()
+      //await Candidate.findByIdAndUpdate(candidateId, { interview: interview})
+    }
+    res.json({ result: response })
```
tko22: let's make responses standardized with the rest of api so it's much easier to interact with it. Like this: https://github.com/hack4impact-uiuc/h4i-recruitment/blob/master/backend/src/api/matchCandidates.js#L84-L93

In this case, result would be empty so don't need to put it in. 

```Javascript
res.json({
    code: 200,
    message: response,
    success: true
})
```

backend/src/api/interview.js
```diff
@@ -0,0 +1,129 @@
...
+  '/:interview_id',
+  errorWrap(async (req, res) => {
+    data = req.body
+    let response = 'Interview Edited Sucessfully'
+    let interviewId = req.params.interview_id
+    let reqSections = data.sections
+    let overallScore = data.overall_score
+    let genNotes = data.general_notes
+
+    if (interviewId == undefined) {
```
tko22: I like how you used `==` instead of `===` when checking `undefined` because `== undefined` also returns True for `null`

backend/src/api/interview.js
```diff
@@ -0,0 +1,129 @@
...
+      )
+    } else if (genNotes != undefined) {
+      Interview.findOneAndUpdate(
+        { _id: new mongodb.ObjectId(interviewId) },
+        { $set: { general_notes: genNotes } },
+        { new: true },
+        function(err, doc) {}
+      )
+    }
+    res.json({ result: response })
```
tko22: same thing as above about responses.

backend/src/api/interview.js
```diff
@@ -0,0 +1,129 @@
...
+  '/',
+  errorWrap(async (req, res) => {
+    let interviews
+    interviews = await Interview.find()
+    res.json({ result: interviews })
+  })
+)
+
+router.get(
+  '/:interview_id',
```
tko22: generally, I would put anything that has a parameter at the end just because frameworks use regex to match urls. since `/:interview_id` is a catch all function for any url with a string after `interviews` (eg `/interviews/10294809570`), say you had another endpoint `/interviews/questions` that you defined **after** this catch all function. Then if you actually tried to get `/interviews/questions`, express would think that you are look for an Interview with id `questions` because it went to the catch all function. 

Not sure if I explained it well, but the following is bad. 
```Javascript
router.get('/:interview_id', (req,res) => {
    console.log(req.params.interview_id)
}
// this function should be written before '/:interview_id'
router.get('/question', (req,res) => {
 .....
}
```

tko22: the way you're doing it right now is fine. just watch out for that, it will save you an hour of debugging.

backend/src/models/candidate.js
```diff
@@ -21,15 +22,7 @@ const Candidate = new mongoose.Schema({
...
-  interviews: [
-    {
-      timeCommitmentNotes: { type: String },
-      dedicationToCommunityNotes: { type: String },
-      techCompetenceNotes: { type: String },
-      otherNotes: { type: String },
-      interviewer: { type: String }
-    }
-  ],
+  //interviews: [Interview],
```
tko22: this should uncommented right?

backend/src/models/interview.js
```diff
@@ -0,0 +1,20 @@
...
+})
+const Interview = new mongoose.Schema({
+  interviewer_key: { type: String },
+  overall_score: { type: Number },
+  general_notes: { type: String },
+  sections: { type: [Section] }
+})
+
+//module.exports =  Interview;
+module.exports = mongoose.model('Interview', Interview)
```
tko22: You may want to export Section and Question as well, just in case.

backend/src/api/interview.js
tko22: Look at #69 for more info

backend/src/api/interview.js
tko22: also once, keys and auth are done, we should also store the the key that submitted this post request into the Interview model.

backend/src/models/interview.js
HRimaw2: thank you Tim, I'm going to edit and make tests

backend/docs/interview_schema.md
```diff
@@ -0,0 +1,248 @@
...
+
+|   id  |   score   | question_text |
+|:------:|:-------:|:------:|
+
+
+## Endpoints Documentation 
+
+
+GET ALL INTERVIEWS: 
+* GET /interviews | interview
```
angadgarg25: This can just be `GET /`. Also is the stuff after the `|` what each endpoint returns?

Issue #82, opened 2018-08-11
add catch all error boundary


tko22:
with nextjs 6 we can add an error boundary in _app.js, which catches errors for all pages

PR #85, opened 2018-08-15
User/tko/stats

tko22:
Ref #82 - some quick styling changes and added stats template and added global error boundary

image

### DIFF
--- a/frontend/src/pages/stats.js
+++ b/frontend/src/pages/stats.js
@@ -1,0 +1,42 @@
+import React, { Component } from 'react'
+import { Container } from 'reactstrap'
+import Head from '../components/head'
+import Nav from '../components/nav'
+
+type Props = {}
+
+class StatsPage extends Component<Props> {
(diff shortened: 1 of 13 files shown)


frontend/src/pages/stats.js
```diff
@@ -0,0 +1,40 @@
...
+          <p># of Sophomores Applied: </p>
+          <p># of Juniors Applied: </p>
+          <p># of Seniors Applied: </p>
+
+          <h3>Matches</h3>
+          <p># of Matches made: </p>
+          <p>Avg number of matches per Candidate</p>
+          <p>Lowest Score: </p>
+          <p>Highest Score: </p>
+          <p>Average Score: </p>
```
angadgarg25: Are there going to be separate endpoints for these?

tko22: nope! i’m thinking of doing everything in the frontend

PR #88, opened 2018-08-18
Added Authentication middleware

### LINKED ISSUE #87: Create middleware that authenticates each endpoint
returns 403 if invalid

tko22:
Resolves #87

**will merge this in after megha gets her frontend patch up first**

Now, every endpoint must give a `key` query. eg `localhost:8080?key=1o2iuo5u95`

Also, made changes to `verify_interviewer` by removing the `:key` part of the `verify_interviewer/:key`  in favor of the http query string. now deployment was modified to support this.

All tests are updated to use this. tests were added to test this middleware

Also fixed a bunch of Frontend issues.

### DIFF
--- a/backend/src/middleware/auth.js
+++ b/backend/src/middleware/auth.js
@@ -1,0 +1,30 @@
+const keyPath =
+  process.env.NODE_ENV === 'test' ? '../../tests/artifacts/test-keys.json' : process.env.KEY_JSON
+const keyData = require(keyPath)
+
+const auth = (req, res, next) => {
+  const key = req.query.key
+  if (key != undefined) {
+    if (key && key.length === 11) {
+      keyVerified = keyData.keys.filter(currKey => currKey.key === key).length !== 0
+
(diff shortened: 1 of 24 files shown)


backend/src/middleware/auth.js
```diff
@@ -0,0 +1,46 @@
...
+
+        // check if there is a corresponding name for the key
+        if (req._key_name == undefined) {
+          msg = `The key '${key}'given doesn't have a corresponding name. Check the json file holding the keys.`
+          console.error(msg)
+          res.status(500).json({
+            code: 500,
+            message: 'Bad Key',
+            success: false
+          })
```
angadgarg25: Right now, I think this would hit the `return next()` line, which we don't want if we can't get the name. Can we `return` after setting the status to 500?

tko22: just fixed it

tko22:
will wait for #103 before merge

PR #108, opened 2018-08-24
User profiles 2 [wip]

CodeBrew28:
Resolves #<Replace me with the issue number>

<Add Screenshots!>


frontend/src/components/candidateCard.js
```diff
@@ -155,6 +156,39 @@ class CandidateCardComponent extends Component {
...
+                  <Badge color="primary">Resume</Badge>
+                </a>
+              ) : (
+                // Needed to create a placeholder class to use display: none,
+                // which causes the element to basically not exist on the page
+                // and take up no space
+                <p className="space-fix-placeholder"> </p>
+              )}
+              <p className="space-fix-placeholder"> </p>
+
```
rgychiu: Since the space-fix-placeholder class has the style display: none; the element will basically take up no space on the page even if you put a space between the paragraph tags or add it between the links. It'll just be treated as if it was never put onto the page or coded in. The 1% margin that you added will give it the space though, so you should be able to remove the space-fix-placeholder paragraph tags that are in between the resume, website, and linkedin links.

PR #109, opened 2018-08-24
Sort facemash elo and number of matches

### LINKED ISSUE #106: Facemash sort
Sort by facemash score on main page and add filters for elo and number of matches to be displayed

rgychiu:
Resolves #106


frontend/src/pages/index.js
```diff
@@ -60,7 +60,7 @@ class HomePage extends Component<Props> {
   }
 
   query = () => {
-    if (this.props.candidates.length == 0) {
+    if (this.props.candidates.length != 0) {
```
angadgarg25: It wasn't querying when we changed the filters so we changed this.

tko22: @CodeBrew28

rgychiu: I dont think you should since if you change it to != 0 that means it wont load in all the candidates. Initially the length of the candidates array is 0 so it'll fetch them all for display (at least to my understanding).

PR #120, opened 2018-08-27
Resolves #116 - changed key storage to sessionStorage

### LINKED ISSUE #116: Store keys in SessionStorage instead of LocalStorage
- [ ] make a helper function that get's the keys from sessionStorage

tko22:
Resolves #116 

uses sessionStorage instead of localStorage


tko22:
thanks @angadgarg25

PR #129, opened 2018-08-28
Tentative candidate styling

rgychiu:
Basic styling of candidate page, hit me with some suggestions

image


tko22:
Looks great! Is there any way to make the right column side bar match up with the ‘back’ button. i believe the candidate fields are aligned left and the back button is aligned right? 

Can we also put the comments in the larger column instead of having it span the entire container?

PR #158, opened 2018-09-04
modified parse sheet - haven't tested yet

### LINKED ISSUE #97: Modify the parser to fit the current candidate application
the parser should create status, year(eg freshman, soph, etc), github contribution, etc. You should be able to validate whether the data parsed and dumped into the database was correct and that nothing went wrong. Pay attention to fields that can be in the wrong format (might want to use enums for certain fields like Graduation Date or Roles Applied and check whether the input matches one of those enums). 

Create a sample response for the current application, export the responses into an excel sheet and test your parser on a separate database.

shreyshrey1:
Resolves #97

### DIFF
--- a/backend/src/api/candidates.js
+++ b/backend/src/api/candidates.js
@@ -57,7 +57,7 @@ router.post(
     // add github to the query as well, althought it isn't
     // a selectable in the frontend
     selectFilters.github = 1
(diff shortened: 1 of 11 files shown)


tko22:
is that it?

PR #201, opened 2018-12-23
Ref #200 Added dev env script + page transitions

### LINKED ISSUE #200: Add shell script that sets up dev environment
Let's use docker to spin up a mongo instance, wiping the previous volume, if there are any. It should then load dummy data then provide instructions to add credentials to `.env` and how to start the backend + frontend

tko22:
Resolves #200 and added page transitions to make it look smoother.

Just to help set up everything. 

`yarn populatedb` will populate the database specified in `.env`

We don't have to use individual dev environments and opt for using a global database that everyone uses (what we did this summer).

### DIFF
--- a/frontend/src/pages/dashboard.js
+++ b/frontend/src/pages/dashboard.js
@@ -92,8 +92,7 @@ class Dashboard extends React.Component<Props> {
       .filter(x => this.state.filters.statuses.includes(x.status))
       .filter(x => this.state.filters.years.includes(x.year))
       .filter(x => !x.role.map(role => this.state.filters.roles.includes(role)).includes(false))
-    console.log(this.state.filters.roles)
-    // .filter(x => x.name.toLowerCase().includes(this.state.search.toLowerCase()))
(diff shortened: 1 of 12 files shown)


frontend/src/pages/dashboard.js
```diff
@@ -92,7 +92,8 @@ class Dashboard extends React.Component<Props> {
       .filter(x => this.state.filters.statuses.includes(x.status))
       .filter(x => this.state.filters.years.includes(x.year))
       .filter(x => !x.role.map(role => this.state.filters.roles.includes(role)).includes(false))
-      .filter(x => x.name.toLowerCase().includes(this.state.search.toLowerCase()))
+    console.log(this.state.filters.roles)
+    // .filter(x => x.name.toLowerCase().includes(this.state.search.toLowerCase()))
```
angadgarg25: do we need to keep this?

tko22: woopsies

tko22: the commented filter shouldn't be uncommented too... just fixed it

PR #208, opened 2018-12-30
Interview rounds initial setup

### LINKED ISSUE #207: Interview Rounds Setup
Sub-issue of #203. Configurations of all rounds are stored in `data/roundData.js`, so that when we move from one round to the next we don't need to delete the previous configuration. Need to implement a way to switch between configurations through the portal by adding a new "director"-type user which has the same privileges as leads but with added post access to a new structure table. This structure table holds the index of the current round in the configuration array in `roundData.js` (as a single row) and may hold more as work on #203 continues. On the frontend, this requires a new page where all users can view configurations but only directors can choose the current round, and the navbar will only display the current round so that only that part of the application can have changing information

arpanlaha:
Resolves #207


backend/src/api/interview.js
```diff
@@ -140,7 +140,7 @@ router.post(
   [leadsOnly],
   errorWrap(async (req, res) => {
     const data = req.body
-    let response = 'Interview Added Sucessfully'
+    let response = 'Interview Added Successfully'
```
tko22: ha nice catch

backend/src/models/structure.js
```diff
@@ -0,0 +1,6 @@
+const mongoose = require('mongoose')
+
+const Structure = new mongoose.Schema({
```
tko22: Maybe add a comment to describe what `Structure` is? Just because Structure is a little vague

backend/src/middleware/auth.js
```diff
@@ -34,11 +35,16 @@ const auth = (req, res, next) => {
...
         // check whether key is a lead's key or a member's key
         // this is used by the leadsOnly middleware
-        if (key.endsWith(leadSuffix)) {
+        if (key.endsWith(leadSuffix) || key.endsWith(directorSuffix)) {
           req._is_lead = true
         } else {
           req._is_lead = false
         }
+        if (key.endsWith(directorSuffix)) {
+          req._is_director = true
```
tko22: line 38's change is smart but I think removing the `|| endsWith(directorSuffx)` from there and doing this might be clearer even if it's a couple more lines of code.
```suggestion
if (key.endsWith(directorSuffix)) {
          req._is_director = true
          req._is_lead = true
} else {
           req._is_director = true
          req._is_lead = true
}
```

frontend/src/components/roundDropdown.js
```diff
@@ -0,0 +1,64 @@
...
+  render() {
+    var round_names = []
+    for (var i = 0; i < roundData.numberRounds; i++) {
+      let curr_round = roundData.rounds[i]
+      let curr_round_label = {
+        value: i,
+        label: curr_round.name.toString()
+      }
+      round_names.push(curr_round_label)
+    }
```
tko22: This is good, no need for change. You could also use `map` here for simplicity but it's just a matter of preference! `map` would be very useful for looping through arrays and generating JSX elements in `render()` though!
```Javascript
roundData.map((round,idx) => {value: idx, label: round.name.toString()})

File: frontend/src/pages/rounds.js [hunk 5]
```diff
@@ -0,0 +1,82 @@
...
+
+  closeModal = () => {
+    this.setState({ modalOpen: false })
+  }
+
+  render() {
+    return (
+      <div className="align-middle round-box">
+        <Modal isOpen={this.state.modalOpen} autoFocus={false}>
+          <ModalBody>User doesn't have the priviliges to change rounds.</ModalBody>
```
tko22: use `&#39;` instead of `'` to remove the red

PR #230, opened 2019-01-20
Interview Rounds Styling

### LINKED ISSUE #229: Interview Rounds Styling
Follow-up to #203 - currently some fields in interview sections are pre-formatted strings instead of react elements, which is ugly. Additionally, notes-only questions are only configurable through roundabout means.

arpanlaha:
Resolves #229


frontend/data/roundData.js
```diff
@@ -14,16 +14,32 @@ const roundData = {
...
+            React.createElement('b', {}, '-2 for each:'),
+            React.createElement('ul', {}, [
+              React.createElement(
+                'li',
+                {},
+                'Hard course-load (still take one point off for each hard class they have as described above)'
+              ),
+              React.createElement('li', {}, 'Too many org obligations')
+            ])
+          ],
```
tko22: does jsx not work here?

arpanlaha: Aryn was trying this syntax earlier so I went with it - I don't think we could get other stuff working

angadgarg25: Yeah for some reason, when we worked with Aria and Roy, this worked while jsx was failing.

PR #252, opened 2019-06-15
Event backend

alicesf2:
Refs #253 

Created models and endpoints for event creation and storage, as well as an attendee model for storing info about event attendees


backend/src/models/attendee.js
```diff
@@ -0,0 +1,12 @@
+const mongoose = require('mongoose')
+
+const Attendee = new mongoose.Schema({
+  name: { type: String, required: true },
+  netid: { type: String, required: true },
+  year: { type: Number },
+  isCandidate: { type: Boolean },
```
angadgarg25: Instead of `isCandidate` I would have this field be the `_id` field from the candidate. If it is null, we can assume that they are not a current candidate.

backend/src/api/events.js
```diff
@@ -0,0 +1,196 @@
...
+router.delete(
+  '/:event_id',
+  errorWrap(async (req, res) => {
+    let response = 'Event Deleted Sucessfully'
+    let id = req.params.event_id
+    const event = await Event.findById(id)
+    if (event === undefined) {
+      response = 'Invalid Delete Event Request'
+    } else {
+      await Event.deleteOne({ _id: new mongodb.ObjectId(id) })
```
angadgarg25: I like all the error checking, but we're using two mongo requests here instead of just one. We can probably just use `deleteOne`. It'll either throw an error in which case we can use a try catch, or it will return an object with attribute `deletedCount` which will be 0. I'm not sure which one, but try doing one of those to reduce the number of queries.

backend/src/models/event.js
```diff
@@ -0,0 +1,14 @@
...
+const { AttendeeSchema } = require('./attendee')
+
+const Event = new mongoose.Schema({
+  name: { type: String },
+  date: { type: String },
+  startTime: { type: Number },
+  endTime: { type: Number },
+  location: { type: String },
+  description: { type: String },
+  attendees: { type: [AttendeeSchema] } // subdocument
```
angadgarg25: make these attributes all required, since you're requiring that every field is present in your endpoint

backend/src/api/events.js
```diff
@@ -0,0 +1,196 @@
...
+    } else {
+      const newEvent = new Event({
+        name: eventName,
+        date: eventDate,
+        startTime: eventStartTime,
+        endTime: eventEndTime,
+        location: eventLocation,
+        description: eventDescription,
+        attendees: []
+      })
```
angadgarg25: Because of the [`errorWrap`](https://github.com/hack4impact-uiuc/h4i-recruitment/blob/master/backend/src/middleware/errorWrap.js) middleware, we don't actually need to account for every possible error case. Instead, we can assume that the input is as we want, and if it isn't, the middleware will catch the exception and send the error message. I can explain more about middleware if you aren't familiar. I forgot to do the system overview of the recruitment tool with you I think, so let me know.

But for this endpoint we can start with constructing this new event by accessing the request data directly. If we make these fields required, the line below this comment will raise an exception, and the middleware will take care of it for us.

backend/src/api/events.js
```diff
@@ -0,0 +1,196 @@
...
+          { new: true },
+          function(err, doc) {}
+        )
+      }
+      if (newDescription != undefined) {
+        Event.findOneAndUpdate(
+          { _id: new mongodb.ObjectId(eventId) },
+          { $set: { description: newDescription } },
+          { new: true },
+          function(err, doc) {}
```
angadgarg25: Let's try to avoid a bunch of separate database calls. We can probably do this only using one interaction with the database.

alicesf2: How would you consolidate into 1 database call? I was just going off of the code in candidates.js and interview.js lol

backend/src/api/events.js
alicesf2: Ooo ok yea could you explain more about middleware? I looked at errorWrap.js but it didn't really help haha. Also would u still need the res.json part for the status code and result or will errorWrap handle that too?

backend/src/api/events.js
angadgarg25: Yeah I figured lol, it does look like this in other places too, but we might as well change it now rather than increasing our tech debt! So with `$set` you should just be able to pass multiple key value pairs in json object (`{a: val1, b: val2}`). Maybe use your conditional statements to set keys to values in your json, if they are not undefined. And then after all the conditionals just issue one mongo request. Ideally, we could make it cleaner (not have `n` conditionals for `n` attributes) but that's not super necessary as long as we can just do 1 mongo request here.

backend/src/api/events.js
angadgarg25: I'd skim [this](https://expressjs.com/en/guide/using-middleware.html) real quick.

Basically in [`app.js`](https://github.com/hack4impact-uiuc/h4i-recruitment/blob/master/backend/src/app.js) it sets up the express app using a bunch of `app.use` statements, which define what happens when an endpoint is called (in the order you want them in). In particular, you can see the routes are used near the bottom (these are the specific endpoints) and then afterwards, an [`errorHandler`](https://github.com/hack4impact-uiuc/h4i-recruitment/blob/master/backend/src/middleware/errorHandler.js) is registered. This middleware simply takes an error and logs it. Basically every middleware is a function of a few things including the request, the response, and the next middleware in the chain. Normally when an error is raised, it just breaks the chain and finishes. With the errorWrap, if the endpoint throws an error, it catches it and call the `next` function, which is the `errorHandler` since that was registered directly after. That's a lot, if its unclear I could explain it on a call, it shouldn't take too long.

BUT, you don't need to worry too much about it for this fix specifically, though it is good to understand in general. All you need to know is that your code is basically wrapped in a try catch loop so that if any errors are raised, they are caught and logged. You should still check for more complex errors manually, but for simple things like a field not being set in the request, in this case you should be able to ignore it and let it raise an error. This assumes that you've made every field in your `Event` model required, otherwise an error won't even be raised.

backend/src/api/events.js
```diff
@@ -0,0 +1,196 @@
...
+// get all events
+router.get(
+  '/',
+  errorWrap(async (req, res) => {
+    let events = await Event.find()
+    res.json({
+      code: 200,
+      message: '',
+      result: events,
+      success: true
```
n3a9: Add trailing comma to all multiline statements in Javascript. Just for reference:

1. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas
2. https://medium.com/@nikgraf/why-you-should-enforce-dangling-commas-for-multiline-statements-d034c98e36f8

backend/src/api/events.js
```diff
@@ -0,0 +1,196 @@
...
+  })
+)
+
+// create a new event
+router.post(
+  '/',
+  errorWrap(async (req, res) => {
+    const data = req.body
+    let response = 'Event Created Successfully'
+    let code = 404
```
n3a9: From what I can tell, these variables (`code`, `success`, `eventName`, ...) are not changed. Use `const` instead of `let`.

backend/src/api/index.js
```diff
@@ -4,12 +4,14 @@ const futureInterview = require('./futureInterview')
 const matchCandidates = require('./matchCandidates')
 const matches = require('./matches')
 const structure = require('./structure')
+const events = require('./events')
```
n3a9: nit: `require`s should be in alphabetical order

backend/src/api/index.js
```diff
@@ -4,12 +4,14 @@ const futureInterview = require('./futureInterview')
...
 
 module.exports = {
   candidates,
   interview,
   futureInterview,
   matchCandidates,
   matches,
-  structure
+  structure,
+  events
```
n3a9: Also alphabetical order

backend/src/api/events.js
angadgarg25: agreed ^

backend/src/api/index.js
angadgarg25: no need for alphabetical order stuff right now. The current code doesn't adhere to it, and it doesn't make sense to address those kind of code quality changes in a feature PR. Doing those kind of changes can happen at a different time, but right now it would lead to a rabbit hole across the whole application.

backend/src/api/events.js
alicesf2: so yarn format gets rid of trailing commas rip we'd have to change that

angadgarg25: true, I forgot thats why we don't have them anywhere

backend/src/api/events.js
```diff
@@ -0,0 +1,137 @@
...
+router.delete(
+  '/:event_id',
+  errorWrap(async (req, res) => {
+    const eventId = req.params.event_id
+    let code = 200
+    let message = 'Event Deleted Successfully'
+    let success = true
+
+    const event = await Event.findByIdAndRemove(eventId)
+    if (event === null) {
```
n3a9: Check any falsey value for an object with `if (!event)`

backend/src/api/events.js
```diff
@@ -0,0 +1,137 @@
...
+      message,
+      success
+    })
+  })
+)
+
+// delete an event
+router.delete(
+  '/:event_id',
+  errorWrap(async (req, res) => {
```
n3a9: I think it is much safer to use conditional setting for the variables, rather than defining them before and then changing their values if the event doesn't exist. For example,

```js
    const eventId = req.params.event_id
    const event = await Event.findByIdAndRemove(eventId)

    const ret = event 
        ? {
            code = 200
            message = 'Event Deleted Successfully'
            success = false
           } 
        : {
            code = 404
            message = 'Event Not Found'
            success = false
          }

    res.json(ret)
```

So in pseudocode, if the event exists assign the json object with the values, otherwise assign it with other values. This would be a good usage for other places as well.

alicesf2:
> beautiful
> 
> 



> beautiful
> 
> 

i dont' get it

tko22:
let's write some tests :) maybe in the next PR

alicesf2:
> let's write some tests :) maybe in the next PR

PR #254, opened 2019-06-18
Resolve security vulnerabilities in dependencies

n3a9:
- Resolve security vulnerabilities in dependencies 
- Remove unused export `getAllCandidates` in [api.js](https://github.com/hack4impact-uiuc/h4i-recruitment/blob/master/frontend/src/utils/api.js)

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -21,7 +21,6 @@
     "csv-parser": "^2.1.0",
     "express": "4.16.3",
     "helmet": "^3.13.0",
-    "https-proxy-agent": "^2.2.0",
     "mongoose": "5.0.12",
     "morgan": "^1.9.0",
     "request": "2.87.0",
--- a/backend/package.json
+++ b/backend/package.json
@@ -21,6 +21,7 @@
     "csv-parser": "^2.1.0",
     "express": "4.16.3",
     "helmet": "^3.13.0",
+    "https-proxy-agent": "^2.2.0",
     "mongoose": "5.0.12",
     "morgan": "^1.9.0",
     "request": "2.87.0",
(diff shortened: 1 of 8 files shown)


n3a9:
> Just curious, why did you change the formatting so much? It was passing formatting checks before so how does this help.

The CI tests for the first two commits were not passing for either build or lint.

1. Lint - https://circleci.com/gh/hack4impact-uiuc/h4i-recruitment/2340
2. Build - https://circleci.com/gh/hack4impact-uiuc/h4i-recruitment/2338

This was straight off of master. Not sure why it was passing before, especially considering that in master we had the undefined `getAllCandidates` export in [api.js](https://github.com/hack4impact-uiuc/h4i-recruitment/blob/master/frontend/src/utils/api.js) that wasn't being caught by CI (should have broken build).

I just ran `yarn format` in the front end directory. Currently if you clone the repository and just run this command then it will still make the same changes.

> Also why did you add a dependency thats not being used?

Hm, didn't realize that it was added and not actually upgraded. Looking at https://github.com/hack4impact-uiuc/h4i-recruitment/network/alerts I saw that there were critical issues with the version that we appeared to be using (1.0.0) so I manually upgraded. Its usage is being detected somewhere, possibly as a dependency of another.

angadgarg25:
Ah okay that's pretty strange. My bad, this is fine then. For the dependency though, I wouldn't add it directly. Maybe figure out what dependency it is required by and update that dependency if possible? Just fix that and you're good to go.

angadgarg25:
It looks like one of our dev dependencies so it's not a huge deal. See if you can update `mockgoose-fix`, but if not thats okay since its a dev dependency.

PR #256, opened 2019-06-19
Fix mongo attributes to be actually required

### LINKED ISSUE #255: MongoDB Schemas Fix
Most of our models, use `require: True` for attributes that we want to ensure exist. However, this does nothing, as it should actually be `required: True`. Let's make this fix and add more robust testing for our models to make sure that these fields can only be used as we expect them to be used.

angadgarg25:
Resolves #255 

Our Mongo Code used `require: True` for required attributes in our schema, when we should really be using `required: True`. This fixes that.


backend/tests/utils.js
```diff
@@ -21,10 +21,10 @@ const createCandidates = async (numOfMatches, elos) => {
   for (let i = 0; i < numOfMatches.length; i++) {
     query.push({
       _id: candidateIds(i),
-      name: '',
+      name: 'a',
```
n3a9: Instead of just 'a' why don't we make this a little more intuitive and make it 'candidate' + index (e.g. 'candidate 5'). Even though the field isn't used for testing right now it can help if we want to make a more robust test for fields down the road and is more straightforward.

caiyingshen: +1, maybe same for major below as well

Issue #258, opened 2019-06-20
Event Overview Page


alicesf2:
Create page that displays all past events and lets users create a new event

PR #259, opened 2019-06-21
Event overview page

alicesf2:
Refs #258 
Screenshot from 2019-06-21 01-08-02
Screenshot from 2019-06-21 01-09-17


frontend/src/components/interviewSectionModular.js
```diff
@@ -15,9 +15,7 @@ class InterviewSectionModular extends Component {
         name="Time Commitment"
         id="time-commitment-input"
       >
-        {options.map(option => (
-          <option value={option.value}>{option.name}</option>
-        ))}
+        {options.map(option => <option value={option.value}>{option.name}</option>)}
```
n3a9: `yarn format`

frontend/src/pages/eventoverview.js
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  handleChange = e => {
+    this.setState({
+      [e.target.name]: e.target.value
+    })
+  }
+
+  handleKeyPress = e => {
+    if (e.charCode === 13) {
```
n3a9: Prefer `event.key === "Enter"` for better readability

frontend/src/pages/eventoverview.js
```diff
@@ -0,0 +1,174 @@
...
+      this.addEvent()
+    }
+  }
+
+  addEvent = async () => {
+    let event = {
```
n3a9: `const`

frontend/src/pages/eventoverview.js
```diff
@@ -0,0 +1,174 @@
...
+      name: this.state.name,
+      date: this.state.date,
+      startTime: this.state.startTime,
+      endTime: this.state.endTime,
+      location: this.state.location,
+      description: this.state.description
+    }
+    await createEvent(event)
```
n3a9: Are there any checks anywhere to make sure that these values have been set? These values are required in the backend but if I leave it empty it will (should) throw an error

frontend/src/pages/eventoverview.js
```diff
@@ -0,0 +1,174 @@
...
+                    <Label>End Time</Label>
+                    <Input name="endTime" placeholder="i.e. 6" onChange={this.handleChange} />
+                  </FormGroup>
+                  <FormGroup>
+                    <Label>Location</Label>
+                    <Input name="location" placeholder="i.e. ECEB" onChange={this.handleChange} />
+                  </FormGroup>
+                  <FormGroup>
+                    <Label>Description</Label>
+                    <Input
```
n3a9: Maybe look at the different input types that are offered: https://reactstrap.github.io/components/form/#Input-Types

Maybe use the time field, text box for the description, etc

frontend/src/pages/eventoverview.js
```diff
@@ -0,0 +1,174 @@
...
+                </Button>
+                <Button color="primary" onClick={this.addEvent}>
+                  Submit
+                </Button>
+              </ModalFooter>
+            </Modal>
+            <Table>
+              <thead>
+                <ActionButton
+                  style={{ marginBottom: 15 }}
```
n3a9: Avoid inline styling if possible

frontend/src/components/interviewSectionModular.js
alicesf2: yea i just updated and reran it lol

frontend/src/pages/eventoverview.js
```diff
@@ -0,0 +1,174 @@
...
+    }
+  }
+
+  openModal = () => {
+    this.setState({ modal: true })
+  }
+
+  closeModal = () => {
+    this.setState({ modal: false })
+  }
```
rgychiu: You should be able to consolidate these two functions (openModal, closeModal) into one function that just changes the modal state with !modal. This way its more concise with just one function that opens and closes the modal and is still readable

rgychiu: That way also the modal will be closed by default on the page and then the only way to open the modal is through one certain button and the only way to close the modal is through another button or two on the modal itself, so the modal state should work out nicely

frontend/src/pages/eventoverview.js
rgychiu: +1 in this case, the marginBottom style can be reused across multiple components as well so it's good to place this in a css/scss/etc. file

rgychiu: Also inline styling is very hard to override if it's ever needed, so putting them in a css file is better for more customizable styling

frontend/src/pages/eventoverview.js
```diff
@@ -0,0 +1,174 @@
...
+                  <th>Start Time</th>
+                  <th>End Time</th>
+                  <th>Location</th>
+                  <th>Attendees</th>
+                  <th>Facebook Event Link</th>
+                  <th />
+                </tr>
+              </thead>
+              <tbody>
+                {this.state.events.map(event => (
```
rgychiu: This map might give a unique key error - in general i think React needs unique keys so that it can perform minimal rendering/rerendering. Disregard if it doesn't, but definitely take a look at this

frontend/src/pages/eventoverview.js
alicesf2: should i put css files in the static folder

angadgarg25: yea, I think we only have one css file and it's in there, so I'd use that.

alicesf2: wait in the actionButton component it looks like i have to pass in the style as props

frontend/src/pages/eventoverview.js
```diff
@@ -0,0 +1,216 @@
...
+      this.setState({
+        alert: 'All fields are required.'
+      })
+    } else {
+      Router.push({ pathname: '/eventoverview' })
+    }
+  }
+
+  render() {
+    let alert
```
n3a9:
```js
const alert = (this.state.alert) ? (<Alert color="danger">{this.state.alert}</Alert>) : null
```

frontend/src/pages/eventoverview.js
```diff
@@ -0,0 +1,216 @@
...
+    if (this.state.alert) {
+      alert = <Alert color="danger">{this.state.alert}</Alert>
+    }
+    return (
+      <>
```
n3a9: Use `<div>` as container wrapper

angadgarg25: @n3a9 is there a downside to using fragments here?

n3a9: Actually yeah I just read more about it and it seems better - for some reason I was always under impression that `<div>` was the move 👍

frontend/src/pages/eventoverview.js
```diff
@@ -0,0 +1,216 @@
...
+        <Nav />
+        <div className="page-content-wrapper">
+          <Container fluid>
+            <Modal isOpen={this.state.modal}>
+              <ModalHeader>Add New Event</ModalHeader>
+              <ModalBody>
+                {alert}
+                <Form>
```
arpanlaha: If you say `<Form onSubmit={this.addEvent}>` and add `type="submit"` to the submit button below, it should allow you to submit the form by hitting enter.

frontend/src/pages/eventoverview.js
```diff
@@ -0,0 +1,216 @@
...
+                      onKeyPress={this.handleKeyPress}
+                    />
+                  </FormGroup>
+                </Form>
+              </ModalBody>
+              <ModalFooter>
+                <Button color="secondary" onClick={this.toggleModal}>
+                  Cancel
+                </Button>
+                <Button color="primary" onClick={this.addEvent}>
```
arpanlaha: `<Button type="submit color="primary" onClick={this.addEvent}>`

alicesf2: rip it doesn't work

n3a9: That's because buttons don't have a `type` prop: https://reactstrap.github.io/components/buttons/

PR #261, opened 2019-06-23
Add the ability to delete interviews

### LINKED ISSUE #242: Ability to delete interviews as a Director
A user with a directors key should have the ability to delete specific interviews, written by other people. This can be done in the workflow of accessing interviews in a candidate's page, as well as the `/interviewlist` page. 

This would be for duplicates, bad interviews, etc.

arpanlaha:
Resolves #242 
Also makes some minor changes to mongoose to get rid of warnings.

### DIFF
--- a/backend/src/api/candidates.js
+++ b/backend/src/api/candidates.js
@@ -1,7 +1,7 @@
 const express = require('express')
 const router = express.Router()
 const mongodb = require('mongodb')
-const { errorWrap, leadsOnly } = require('../middleware')
+const { errorWrap, leadsOnly, directorsOnly } = require('../middleware')
 const { Candidate, Comment, Interview } = require('../models')
 const {
   statusEnum,
@@ -296,16 +296,16 @@ router.put(
 )
 router.delete(
   '/:candidateId/interviews/:interviewId',
-  [leadsOnly],
+  [directorsOnly],
   errorWrap(async (req, res) => {
     let response = 'Interview Deleted Sucessfully'
     const candidate = await Candidate.findById(req.params.candidateId)
-    await candidate.interviews.id(req.params.interviewId).remove()
+    const result = await candidate.interviews.id(req.params.interviewId).remove()
     await candidate.save() // save cascades down to subdocument
     res.json({
       code: 200,
       message: response,
-      result: {},
+      result: result,
       success: true
     })
   })
--- a/backend/src/api/candidates.js
+++ b/backend/src/api/candidates.js
@@ -298,14 +298,14 @@ router.delete(
   '/:candidateId/interviews/:interviewId',
   [directorsOnly],
   errorWrap(async (req, res) => {
-    let response = 'Interview Deleted Sucessfully'
+    const response = 'Interview Deleted Sucessfully'
     const candidate = await Candidate.findById(req.params.candidateId)
-    const result = await candidate.interviews.id(req.params.interviewId).remove()
+    await candidate.interviews.id(req.params.interviewId).remove()
     await candidate.save() // save cascades down to subdocument
     res.json({
       code: 200,
       message: response,
-      result: result,
+      result: {},
       success: true
     })
   })
--- a/backend/src/api/interview.js
+++ b/backend/src/api/interview.js
@@ -128,7 +128,7 @@ router.get(
   '/:interview_id',
   [leadsOnly],
   errorWrap(async (req, res) => {
-    const retInterview = await Interview.findById(new mongodb.ObjectId(req.params.interview_id))
+    const retInterview = await Interview.findById(req.params.interview_id)
     res.json({
       code: 200,
       message: '',
--- a/backend/src/api/interview.js
+++ b/backend/src/api/interview.js
@@ -128,23 +128,11 @@ router.get(
   '/:interview_id',
   [leadsOnly],
   errorWrap(async (req, res) => {
-    const retInterview = await Interview.findById(req.params.interview_id)
+    const retInterview = await Interview.findById(new mongodb.ObjectId(req.params.interview_id))
     res.json({
       code: 200,
       message: '',
       result: retInterview,
-      success: true
-    })
-  })
-)
-router.get(
-  '/:interview_id',
-  errorWrap(async (req, res) => {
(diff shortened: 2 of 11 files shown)


backend/src/api/interview.js
```diff
@@ -137,18 +137,6 @@ router.get(
     })
   })
 )
-router.get(
```
arpanlaha: duplicate route (how did this happen lol)

backend/src/index.js
```diff
@@ -3,7 +3,10 @@ const app = require('./app')
 const mongoose = require('mongoose')
 
 // connect to mongoose
-mongoose.connect(process.env.MONGO_URL)
+mongoose.connect(
+  process.env.MONGO_URL,
+  { useNewUrlParser: true }
```
arpanlaha: got rid of a warning for me

backend/src/api/candidates.js
```diff
@@ -296,16 +296,16 @@ router.put(
 )
 router.delete(
   '/:candidateId/interviews/:interviewId',
-  [leadsOnly],
+  [directorsOnly],
   errorWrap(async (req, res) => {
     let response = 'Interview Deleted Sucessfully'
     const candidate = await Candidate.findById(req.params.candidateId)
-    await candidate.interviews.id(req.params.interviewId).remove()
+    const result = await candidate.interviews.id(req.params.interviewId).remove()
```
arpanlaha: return what u deleted (should I not do this?)

frontend/src/components/candidates/candidateInterviewsModal.js
```diff
@@ -4,16 +4,28 @@ import React from 'react'
 import { Container, Modal, ModalHeader, ModalBody, ModalFooter, Button } from 'reactstrap'
 import InterviewCard from '../interviewCard'
 import InterviewDetails from '../interviewDetails'
+import { getCandidateInterviews } from '../../utils/api'
```
arpanlaha: instead of passing interviews in as props, pass the ID in and fetch on `componentDidMount`
this makes it easier to update the interviews displayed in the component as its local to the component itself, otherwise I'd have to do redux and I don't really want to do that

frontend/src/components/candidates/candidateInterviewsModal.js
```diff
@@ -26,6 +38,32 @@ class CandidateInterviewsModal extends React.Component<Props> {
       currentInterview: null
     })
   }
+  handleDeleteClick = (interviewId, candidateId) => {
```
arpanlaha: taken from `interviewPortal`

frontend/src/components/interviewCard.js
```diff
@@ -17,29 +17,57 @@ type Props = {
...
-          </div>
-        </CardBody>
-      </Card>
+      <>
+        <Modal isOpen={this.state.modal} toggle={this.toggle}>
+          <ModalHeader>
+            Are you sure you want to delete this interview? You won't be able to get it back.
+          </ModalHeader>
+          <ModalFooter>
+            <Button onCLick={this.toggle}>Return</Button>
```
arpanlaha: @rgychiu css pls
image

backend/src/api/candidates.js
angadgarg25: this is fine, I don't think we rely on DELETE endpoints returning the deleted objects anywhere, though that is a common pattern

frontend/src/components/interviewCard.js
```diff
@@ -17,29 +17,57 @@ type Props = {
...
+  async handleDeleteClick() {
+    await this.props.handleDelete()
+    this.toggle()
+  }
+
+  toggle() {
+    this.setState({
+      modal: !this.state.modal
+    })
+  }
```
angadgarg25: lets use anonymous function syntax to avoid bind here?

backend/src/api/candidates.js
alan-fang: I think the standard practice is to just leave it out - the way to verify that a delete to occur would be to search in the database or see a UI refresh

frontend/src/components/interviewCard.js
alan-fang: Perhaps conditionally rendering the delete button based on who is viewing the component might be better. Not sure the feasibility of this though

tko22: `margin-right`

frontend/src/components/candidates/candidateInterviewsModal.js
```diff
@@ -26,6 +36,23 @@ class CandidateInterviewsModal extends React.Component<Props> {
       currentInterview: null
     })
   }
+  delete = async interview => {
+    this.setState({
+      loading: true
```
alan-fang: cool, is there some kind of spinner icon that shows during this state?

backend/src/api/candidates.js
```diff
@@ -296,16 +296,16 @@ router.put(
 )
 router.delete(
   '/:candidateId/interviews/:interviewId',
-  [leadsOnly],
+  [directorsOnly],
   errorWrap(async (req, res) => {
     let response = 'Interview Deleted Sucessfully'
```
alan-fang: hmmm unless I am missing some context here, the response never changes. Should it be changed to a `const` instead?

arpanlaha: fixed, will update other ones also

frontend/src/components/candidates/candidateInterviewsModal.js
arpanlaha: dont think so, this was copied from elsewhere
i'll probably get rid fo this

arpanlaha: this was copied from a different component , forgot to take loading out

frontend/src/components/interviewCard.js
arpanlaha: with the current pre-auth implementation I dont think its possible

arpanlaha: but once thats done it should be

arpanlaha: @tko22 still looks the same

alan-fang:
lgtm

PR #269, opened 2019-06-29
Conditional rendering in navigation based on user role

### LINKED ISSUE #89: Make Navbar show different tabs dependent on the user's level
- if the user is a member, then it shouldn't show the dashboard or stats board, only just the main page and facesmash
- if user is lead, it should show everything

n3a9:
Resolves #89 

## Summary

Allows for conditional rendering based on user role. Specific navigation tabs can be easily changed to be viewed by members.

## Implementation

Capture the `is_lead` field of result of login request and set state `isLead` boolean. If true, then show all navigation tabs in the nav bar.

Also includes:
- Change `isOpen` to `showModal` state variable name for clearer meaning
- Prefer inline && operator for conditional rendering (https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator)

## Testing

Ensure that login from both home page and the modal allows for this change to take effect.


frontend/src/components/nav.js
```diff
@@ -39,44 +39,56 @@ class NavigationBar extends Component {
...
     })
     const { success, result } = await validateKey(this.state.currentKey)
     if (success) {
       localStorage.setItem('interviewerKey', this.state.currentKey)
+      localStorage.setItem('interviewerRoleIsLead', result.is_lead)
       localStorage.setItem('interviewerName', result.name)
       Router.push('/dashboard')
+
+      this.setState({
+        isLead: result.is_lead === 'true',
```
alan-fang: Is this robust? It seems like I could just go to my browser console and type in `window.localStorage.setItem('interviewerRoleIsLead', true)` and manually set my own permissions level.

frontend/src/components/nav.js
```diff
@@ -39,44 +39,56 @@ class NavigationBar extends Component {
   constructor(props) {
     super(props)
     this.state = {
-      isOpen: false,
+      isLead: false,
```
alan-fang: How many levels of user roles are there? If there are more than 2 (or there plan to be), then I'd recommend using an enum to prevent bloat

frontend/src/components/nav.js
n3a9: No, this won't set your permission level. All this will do is show you the other nav links, but won't change your actual permission in terms of functionality.

frontend/src/components/nav.js
n3a9: The number of roles doesn't matter here - roles are defined to be 'lead' or 'not lead'. As described in the issue, we should only show the nav components when the user is a lead (which would mean the role could be a lead or a director).

frontend/src/components/nav.js
alan-fang: I think that's still bad practice, but probably not a big issue wrt the scope of what this portal does ¯\_(ツ)_/¯

n3a9: I agree that there's probably a better and more secure way to implement this - I will look into it. My intuition was that members won't edit local storage just to see pages in the nav that they won't have access to anyways.

n3a9: Changed implementation so it prevents local storage.

Issue #277, opened 2019-07-04
Attendee Backend


alicesf2:
Write endpoints for modifying event attendee data

PR #278, opened 2019-07-04
Attendee Backend

alicesf2:
Refs #277

### DIFF
--- a/backend/src/api/attendees.js
+++ b/backend/src/api/attendees.js
@@ -22,7 +22,7 @@ router.get(
   '/:netId',
   errorWrap(async (req, res) => {
     const netId = req.params.netId
-    const attendee = await Attendee.find({netId})
+    const attendee = await Attendee.find({ netId })
     res.json({
       code: 200,
       result: attendee,
@@ -37,10 +37,10 @@ router.post(
   errorWrap(async (req, res) => {
     const data = req.body
     const newAttendee = new Attendee({
-        name: data.name,
-        netId: data.netId,
-        email: data.netId + "@illinois.edu",
-        year: data.year
+      name: data.name,
+      netId: data.netId,
+      email: data.netId + '@illinois.edu',
+      year: data.year
     })
 
     await newAttendee.save()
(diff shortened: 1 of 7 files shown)


backend/src/api/attendees.js
```diff
@@ -0,0 +1,128 @@
...
+    const attendee = await Attendee.findOneAndUpdate(
+      { netId },
+      { $set: fieldsToUpdate },
+      { new: true }
+    )
+    const ret = attendee
+      ? {
+          code: 200,
+          message: 'Attendee Updated Successfully',
+          success: true
```
alan-fang: A random comment that is not necessarily related to your PR.. is there a reason why we have a `code` and a `success` field? Wouldn't a `code` field of 200 suggest that the response is a success? Or am I misinterpreting the meaning of `success` here?

backend/src/models/event.js
```diff
@@ -8,7 +8,8 @@ const Event = new mongoose.Schema({
   endTime: { type: String, required: true },
   location: { type: String, required: true },
   description: { type: String, required: true },
-  attendees: { type: [AttendeeSchema], required: true }, // subdocument
+  attendees: { type: [mongoose.Schema.Types.ObjectId], required: true },
+  lateAttendees: { type: [mongoose.Schema.Types.ObjectId], required: true },
```
alan-fang: omg lol

backend/src/api/attendees.js
```diff
@@ -0,0 +1,128 @@
...
+
+// create a new attendee
+router.post(
+  '/',
+  errorWrap(async (req, res) => {
+    const data = req.body
+    const newAttendee = new Attendee({
+      name: data.name,
+      netId: data.netId,
+      email: data.netId + '@illinois.edu',
```
alan-fang: hmmmmm the functionality I'm working on (Workspaces) is supposed to allow us to give the tool to other student groups and schools for them to use. So it's not necessarily true that the email should be suffixed with '@illinois.edu'. Similarly, the field 'netId' probably is not a great descriptor either (although this is a non-issue and more of a naming nit). I recommend dropping the email column entirely or having emails be an explicit entry.

alicesf2: ok

backend/src/models/event.js
alicesf2: ok

tko22:
📸🐣

PR #283, opened 2019-07-07
Workspace endpoint

### LINKED ISSUE #279: Create endpoint for workspaces
Part of #249 
This is the start of how we're going to allow directors to change recruitment cycles more seamlessly. These endpoints will be under `cycle`.

chyku:
Endpoints to hit to create and get recruitment cycles.

Note: Why is frontend-lint failing? I only did yarn format on the frontend then it started failing..

Resolves #279

Screenshots of the endpoints:
image
image
image


backend/src/models/cycle.js
```diff
@@ -0,0 +1,14 @@
+const mongoose = require('mongoose')
+
+/**
+ * Holds information about a recruitment cycle; attributes are subject to change.
+ *
+ * term: (for example) SP19, FA20
+ * chapter: (or school) UIUC, GaTech
```
alanfang7: hmmm, I don't think we should be using abbreviations as chapters. One of my biggest annoyances looking at past spreadsheets and doing recruiting for other RSOs is that there is no standardization of naming (ie someone will write CompE and someone else will write CE and someone else will write Computer Engineering). Maybe change this comment to say `Full name of chapter (or school)` and write out `Georgia Tech not GaTech`

backend/src/models/cycle.js
```diff
@@ -0,0 +1,14 @@
...
+ */
+
+const Cycle = new mongoose.Schema({
+  term: String,
```
alanfang7: I think these are both required fields and should thus have the `required` flag set to `true`

frontend/src/pages/interviewlist.js
```diff
@@ -98,15 +98,16 @@ class InterviewListPage extends React.Component<Props> {
...
-                          {this.state.interviewingInterviews.map(interview =>
-                            interviewGetCategorySection(interview) !== null &&
-                            interviewGetCategorySection(interview).response.text === category ? (
-                              <li>
-                                <Link href={`/candidate?id=${interview.candidate_id}`}>
-                                  {interview.candidate_name}
-                                </Link>
-                              </li>
-                            ) : null
+                          {this.state.interviewingInterviews.map(
```
alanfang7: hmmmm I noticed that you changed frontend files and now lint is failing. Did you reformat the frontend manually? I think @angadgarg25 suggests that we just leave formatting up to `yarn`, so if you made any changes just change them back. Let me know if my assumption is incorrect though!

backend/tests/cycle.test.js
```diff
@@ -0,0 +1,56 @@
...
+const app = require('../src/app')
+const { Cycle } = require('../src/models')
+const { KEY } = require('./utils.js')
+require('./mongo_utils')
+
+beforeEach(async () => {
+  await Cycle.deleteMany()
+})
+
+describe('App can run', done => {
```
alanfang7: wow nice job with these tests!

backend/src/api/cycle.js
```diff
@@ -0,0 +1,69 @@
+const express = require('express')
+const { Cycle } = require('../models')
+const { directorsOnly, errorWrap } = require('../middleware')
+const router = express.Router()
+
+// Get all cycles (previous and current)
+router.get(
```
alanfang7: I think we should probably just put this call behind a `[directorsOnly]` security authorization

backend/src/api/cycle.js
```diff
@@ -0,0 +1,69 @@
...
+  })
+)
+
+// Create a new cycle
+router.post(
+  '/',
+  [directorsOnly],
+  errorWrap(async (req, res) => {
+    let response = 'Cycle Created Successfully'
+    let code = 404
```
alanfang7: yeah 400 not 404, 404 is `not found`

backend/src/api/candidates.js
```diff
@@ -13,6 +13,7 @@ const {
 } = require('../utils/enums')
 const { getGithubContributions } = require('../utils/gitScraper')
 
+// Get all candidates
```
alanfang7: Thanks for adding all of this documentation 😁

frontend/src/pages/interviewlist.js
alan-fang: ohhh hmmm just read your note.. pinging @angadgarg25 to see what's going wrong here

backend/tests/cycle.test.js
```diff
@@ -0,0 +1,56 @@
...
+      .expect(200)
+    expect(res.body.result.term).to.eq('FA19')
+  })
+})
+
+describe('POST /cycle', () => {
+  it('should create a cycle', async () => {
+    const cycle = new Cycle({
+      term: 'FA19',
+      chapter: 'UIUC'
```
alanfang7: Correspondingly, `University of Illinois at Urbana-Champaign` not `UIUC`. lmk if you have a strong pushback against this though

backend/tests/cycle.test.js
```diff
@@ -0,0 +1,56 @@
...
+      .expect(200)
+    expect(res.body.result).to.be.an('array')
+  })
+})
+
+describe('GET /cycle/:cycleId', () => {
+  it('should get a cycle by ID', async () => {
+    const cycle = new Cycle({
+      term: 'FA19',
+      chapter: 'UIUC'
```
alan-fang: Correspondingly, University of Illinois at Urbana-Champaign not UIUC. lmk if you have a strong pushback against this though

angadgarg25:
Probably the lint on circle CI is newer. If you installed yarn via brew run `brew upgrade yarn` otherwise run `npm upgrade --global yarn`. Then just run `yarn` again, and then `yarn format`.

frontend/src/pages/interviewlist.js
angadgarg25: try out my comment below

backend/src/models/cycle.js
angadgarg25: agreed

backend/src/api/cycle.js
```diff
@@ -0,0 +1,69 @@
...
+      response = 'Invalid term'
+    }
+    if (!newChapter) {
+      response = 'Invalid chapter'
+    }
+
+    const cycle = new Cycle({
+      term: newTerm,
+      chapter: newChapter
+    })
```
angadgarg25: This cycle is created even if the term or chapter is invalid. Maybe consider using `else if` and `else` to check if chapter is invalid and create the cycle respectively. You could also leave it as is and just rely on `cycle.save()` throwing an error, but that won't happen right now either since you haven't required those fields.

backend/tests/cycle.test.js
chyku: nah no prob, it makes sense!

backend/src/api/cycle.js
chyku: Thanks for the feedback!! so as I understand, if those fields are required then this part will be fine?

chyku:
> Probably the lint on circle CI is newer. If you installed yarn via brew run `brew upgrade yarn` otherwise run `npm upgrade --global yarn`. Then just run `yarn` again, and then `yarn format`.

This worked, thank you!!

backend/src/api/cycle.js
alan-fang: I'm not sure that this would be fine (think the await cycle.save() will throw an error but the message you want won't be surfaced correctly), but we can text/fix in a follow up revision

Issue #280, opened 2019-07-07
Create frontend interface for admins to change workspaces


chyku:
Part of #249
Create a form with required fields for admins to change and create workspaces (or cycles).

PR #297, opened 2019-07-16
Director only nav items, workspace schema

alan-fang:
Part of #280 

Director only nav items, and workspace model

The "workspaces" nav item is not available in leads or members nav bar

Screen Shot 2019-07-15 at 10 01 22 PM

### DIFF
--- a/backend/src/models/workspace.js
+++ b/backend/src/models/workspace.js
@@ -1,0 +1,14 @@
+const mongoose = require('mongoose')
+
+/**
+ * Partitions existing recruitment information into groups (aka "workspaces").
+ *
+ * name: name of the workspace. Functionally also a unique identifier.
+ * creator: name of the director who created the workspace
+ */
+
+const Workspace = new mongoose.Schema({
+  name: { type: String, required: true, unique: true },
+  creator: { type: String, required: true }
+})
+module.exports = mongoose.model('Workspace', Workspace)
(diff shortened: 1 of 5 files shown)


frontend/src/components/nav.js
```diff
@@ -133,6 +134,15 @@ class NavigationBar extends Component {
...
                 </Link>
               </NavItem>
+              {this.state.isDirector && (
+                <>
+                  <NavItem>
+                    <Link prefetch href="/workspaces">
+                      <a className="nav-bar-link pl-3">Workspaces</a>
+                    </Link>
+                  </NavItem>
+                </>
```
angadgarg25: There's only one top level component here so I don't think we need the Fragment? Unless theres something I missing.

backend/src/models/workspace.js
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Partitions existing recruitment information into groups (aka "workspaces").
+ *
+ * name: name of the workspace. Functionally also a unique identifier.
+ * creator: name of the director who created the workspace
+ */
+
+const Workspace = new mongoose.Schema({
+  name: { type: String, required: true, unique: true },
```
angadgarg25: since this is user created, how will we make sure it unique on the frontend?

alan-fang: 🤔🤔🤔 will address this in frontend and API

Issue #249, opened 2019-05-27
Add Workspace Functionality


rgychiu:
Currently, the process for managing the recruitment portal between each recruitment cycle involves creating a new database along with another script that parses an excel file of applicants into the portal. We want to simplify and automate this task such that admins can create another “round” of applications while minimizing the amount of manual work done (like creating a new database). We also want to make the parsing of uploaded excel files relatively general to allow for different candidate/application structures to be supported. We’d also like for old data from previous recruitment cycles to be kept so that we’ll be able to use the data for more insight or other purposes. These rounds would identify a unique batch of applicants and a unique school. Once auth-infra is setup, co-directors should be able to turn on access to these rounds to specific members.

In the frontend, this will most likely include an admin-level button or other control feature that will clear current candidates and prepare for parsing more candidates. This will also most likely require collaboration with the auth-infra pair to ensure that only admins can access this feature.

In the backend, this feature may require interacting with the database to create a new collection for each new recruitment cycle, or some other way of storing the current recruitment cycle and creating a new cycle. It may also need to check for other issues like starting a new round of recruitment twice without any data being inputted (ie. admin clears the current round, then clears again).

PR #298, opened 2019-07-16
refactor chapter -> workspaceName

alan-fang:
part of #249 

Using the workspace name instead of the chapter (which functionally represent the same thing)

### DIFF
--- a/backend/tests/cycle.test.js
+++ b/backend/tests/cycle.test.js
@@ -30,7 +30,7 @@ describe('GET /cycle/:cycleId', () => {
   it('should get a cycle by ID', async () => {
     const cycle = new Cycle({
       term: 'FA19',
-      chapter: 'University of Illinois at Urbana-Champaign'
+      workspaceName: 'University of Illinois at Urbana-Champaign'
     })
     await cycle.save()
     const res = await request(app)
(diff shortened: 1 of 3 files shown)


backend/tests/cycle.test.js
```diff
@@ -30,7 +30,7 @@ describe('GET /cycle/:cycleId', () => {
   it('should get a cycle by ID', async () => {
     const cycle = new Cycle({
       term: 'FA19',
-      chapter: 'University of Illinois at Urbana-Champaign'
+      workspaceName: 'University of Illinois at Urbana-Champaign'
```
angadgarg25: maybe just add Hack4Impact to the end of the string so that we're following our own example here.

backend/tests/cycle.test.js
```diff
@@ -44,7 +44,7 @@ describe('POST /cycle', () => {
   it('should create a cycle', async () => {
     const cycle = new Cycle({
       term: 'FA19',
-      chapter: 'University of Illinois at Urbana-Champaign'
+      workspaceName: 'University of Illinois at Urbana-Champaign'
```
angadgarg25: same here

alan-fang:
nah it's fine! just a small naming change to make things consistent. To me, cycles and seasons are pretty much the same thing

Issue #299, opened 2019-07-17
Link attendee name to candidate page


alicesf2:
For single event page, display attendees and link their name to their candidate profile page if they're a candidate

PR #300, opened 2019-07-17
Display event attendees and link them to candidate page

alicesf2:
Refs #299 
Screen Shot 2019-07-17 at 12 54 29 AM

### DIFF
--- a/backend/src/api/attendees.js
+++ b/backend/src/api/attendees.js
@@ -1,5 +1,6 @@
 const express = require('express')
 const mongodb = require('mongodb')
+const mongoose = require('mongoose')
 const router = express.Router()
 const { errorWrap } = require('../middleware')
 const { Attendee, Candidate } = require('../models')
@@ -37,12 +38,12 @@ router.post(
   errorWrap(async (req, res) => {
     const data = req.body
     const candidate = await Candidate.findOne({email: data.email})
-    const isCandidate = candidate ? true : false
+    const candidateId = candidate ? candidate._id : null
     const newAttendee = new Attendee({
       name: data.name,
       email: data.email,
       year: data.year,
-      isCandidate
+      candidateId
     })
 
     await newAttendee.save()
--- a/backend/src/api/attendees.js
+++ b/backend/src/api/attendees.js
@@ -36,10 +36,13 @@ router.post(
   '/',
   errorWrap(async (req, res) => {
     const data = req.body
+    const candidate = await Candidate.findOne({email: data.email})
+    const isCandidate = candidate ? true : false
     const newAttendee = new Attendee({
       name: data.name,
       email: data.email,
-      year: data.year
+      year: data.year,
+      isCandidate
     })
 
     await newAttendee.save()
--- a/backend/src/api/attendees.js
+++ b/backend/src/api/attendees.js
@@ -37,7 +37,7 @@ router.post(
   '/',
   errorWrap(async (req, res) => {
     const data = req.body
-    const candidate = await Candidate.findOne({email: data.email})
+    const candidate = await Candidate.findOne({ email: data.email })
     const candidateId = candidate ? candidate._id : null
     const newAttendee = new Attendee({
       name: data.name,
--- a/backend/src/api/attendees.js
+++ b/backend/src/api/attendees.js
@@ -1,6 +1,4 @@
 const express = require('express')
-const mongodb = require('mongodb')
-const mongoose = require('mongoose')
 const router = express.Router()
 const { errorWrap } = require('../middleware')
 const { Attendee, Candidate } = require('../models')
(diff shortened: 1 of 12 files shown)


backend/src/api/attendees.js
```diff
@@ -1,5 +1,6 @@
 const express = require('express')
 const mongodb = require('mongodb')
+const mongoose = require('mongoose')
```
angadgarg25: unused import?

frontend/src/pages/event.js
```diff
@@ -0,0 +1,114 @@
...
+                        <td>{attendee.year}</td>
+                      </tr>
+                    ) : (
+                      <tr key={attendee._id}>
+                        <td>{attendee.name}</td>
+                        <td>{attendee.email}</td>
+                        <td>{attendee.year}</td>
+                      </tr>
+                    )
+                  )}
```
angadgarg25: Actually we should bring this `attendee.candidateId ?` down to just the name element, so that we don't have to rewrite email and year.

backend/src/api/attendees.js
```diff
@@ -36,10 +37,13 @@ router.post(
...
   errorWrap(async (req, res) => {
     const data = req.body
+    const candidate = await Candidate.findOne({ email: data.email })
+    const candidateId = candidate ? candidate._id : null
     const newAttendee = new Attendee({
       name: data.name,
       email: data.email,
-      year: data.year
+      year: data.year,
+      candidateId
```
chyku: do you have to do `candidateId: candidateId` or is it filled in automatically by mongoose?

backend/src/api/events.js
```diff
@@ -117,20 +146,24 @@ router.put(
...
+        const candidate = await Candidate.findOne({ email: data.email })
+        const candidateId = candidate ? candidate._id : null
+
         attendee = new Attendee({
           name: data.name,
-          email: attendeeEmail,
-          year: data.year
+          email: data.email,
+          year: data.year,
+          candidateId
```
chyku: same question as above

frontend/src/pages/event.js
```diff
@@ -0,0 +1,114 @@
+import React, { Component } from 'react'
+import { Table, Row, Button, Col, FormGroup, Label, Input, Container } from 'reactstrap'
+import ActionButton from '../components/actionButton'
+import Nav from '../components/nav'
+import Head from '../components/head'
+import { getEventById, getEventAttendees } from '../utils/api'
+import Router from 'next/router'
+import Link from 'next/link'
+
+class Event extends React.Component<Props> {
```
arpanlaha: Tiny thing but could just say `Component` here since it's imported, or otherwise remove `Component` from the imports

frontend/src/pages/event.js
```diff
@@ -0,0 +1,114 @@
...
+      attendeeEmails: [],
+      attendees: []
+    }
+  }
+
+  async componentDidMount() {
+    const { query } = Router
+    const { result } = await getEventById(query.id)
+    this.setState({
+      name: result.name,
```
arpanlaha: this could break if the request fails

arpanlaha: could probably just say
```js
result && this.setState({...})
```

frontend/src/pages/event.js
```diff
@@ -0,0 +1,114 @@
...
+      endTime: result.endTime,
+      location: result.location,
+      description: result.description,
+      fbLink: result.fbLink,
+      attendeeEmails: result.attendeeEmails
+    })
+
+    const res = await getEventAttendees(query.id)
+    this.setState({
+      attendees: res.result
```
arpanlaha: same as above

backend/src/api/attendees.js
angadgarg25: Yeah, with javascript this is equivalent to putting `candidateId: candidateId`! It's not a mongo thing, just a javascript thing.

frontend/src/pages/event.js
angadgarg25: Double check on this before you fix alice. It might just assign everything to undefined, since I believe failing requests still return a result (with status code and error message, etc.). Not sure tho, maybe it would break.

backend/src/api/attendees.js
chyku: javascript is wild

frontend/src/pages/event.js
tko22: it will break

tko22: why we do this https://github.com/hack4impact-uiuc/h4i-recruitment/blob/ea5ad18eb4e637ca940228cc540ca228b0c7a022/frontend/src/pages/candidate.js#L53

frontend/src/pages/event.js
alan-fang: Based on @n3a9 's refactor I'd say just extend `Component`

PR #302, opened 2019-07-18
Code Sanitization

n3a9:
**Includes**
- [x] Standardizing named imports for all `extends Component`
- [x] Move towards [inline if component rendering](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator) rather than ternary operators ending with `: <></>`
- [x] Standardizing whitespacing between functions
- [x] Removing unused imports
- [x]  Markdown linting

### DIFF
--- a/frontend/src/components/candidateCard.js
+++ b/frontend/src/components/candidateCard.js
@@ -3,7 +3,7 @@ import React, { Component } from 'react'
 import { connect } from 'react-redux'
 import { bindActionCreators } from 'redux'
 import Link from 'next/link'
-import { Card, CardBody, CardTitle, Button, Badge, Row, Col } from 'reactstrap'
+import { Card, CardBody, CardTitle, Badge, Row, Col } from 'reactstrap'
 import Router from 'next/router'
 import { setCandidateStatus } from '../utils/api'
 import { statusEnum } from '../utils/enums'
@@ -13,17 +13,13 @@ import CandidateLinksBadge from '../components/candidateLinksBadge'
 const handler = (_id: string) =>
   Router.push({
     pathname: '/candidate',
-    query: { id: _id }
+    query: { id: _id },
   })
-
-type Props = {
-  candidate: Array<mixed> // TODO: make this more specific
-}
 
 const mapDispatchToProps = dispatch => {
   return bindActionCreators(
     {
-      setStatus
+      setStatus,
     },
     dispatch
   )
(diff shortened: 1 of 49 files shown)


frontend/src/components/candidateDropdown.js
```diff
@@ -26,26 +31,23 @@ const mapDispatchToProps = dispatch => {
...
   state = {
     isClearable: false,
     isSearchable: true,
     selectedOption: null
   }
+
+  constructor(props) {
+    super(props)
+  }
+
```
arpanlaha: jw, was there a reason this was moved?

frontend/src/components/candidateCard.js
```diff
@@ -49,107 +47,83 @@ class CandidateCardComponent extends Component {
...
                   // Replaced the span with Badge for easier, more consistent styling
                   <h5>
                     <Badge color="success">Accepted</Badge>
                   </h5>
-                ) : (
-                  <></>
                 )}
 
-                {candidate.status && candidate.status === 'Pending' ? (
+                {candidate.status && candidate.status === 'Pending' && (
```
alan-fang: thanks for doing this!

frontend/src/components/candidateDropdown.js
```diff
@@ -60,24 +62,26 @@ class CandidateDropdown extends Component<Props, State> {
...
     var candidate_names = []
     for (var i = 0; i < candidates.length; i++) {
       let curr_cand = this.props.candidates[i]
       let curr_name_label = { value: curr_cand.name.toString(), label: curr_cand.name.toString() }
       candidate_names.push(curr_name_label)
     }
+
     return (
-      <Fragment>
+      <>
```
alan-fang: nit: do we need a fragment if it only has one component?

frontend/src/components/candidateDropdown.js
n3a9: I thought that initial state declarations should be the first thing in the component class, so that its easier to track the variables. Let me know what you think!

frontend/src/components/candidateDropdown.js
n3a9: Nope - good catch! Will remove.

frontend/src/components/candidateDropdown.js
arpanlaha: no real preference on my end as long as its consistent (which it seems to be), so sounds good!

PR #303, opened 2019-07-18
Proposal: Add trailing commas to lint

n3a9:
**Proposal**

Add `es5`[trailing commas](https://prettier.io/docs/en/options.html#trailing-commas) for [cleaner diffs](https://medium.com/@nikgraf/why-you-should-enforce-dangling-commas-for-multiline-statements-d034c98e36f8).

### DIFF
--- a/frontend/.prettierrc.json
+++ b/frontend/.prettierrc.json
@@ -1,5 +1,6 @@
 {
   "semi": false,
   "singleQuote": true,
-  "printWidth": 100
+  "printWidth": 100,
+  "trailingComma": "es5"
 }
--- a/frontend/src/actions/actionCreators.js
+++ b/frontend/src/actions/actionCreators.js
@@ -12,22 +12,22 @@ import {
   RESET_FILTERS,
   SET_ROUND,
   SET_SELECTED_ROUND,
-  SET_VALID_FORMAT
+  SET_VALID_FORMAT,
 } from './actionTypes.js'
 import { getCandidates } from '../utils/api'
 // Action Creators
 
 export const setRoundRedux = round => ({
   type: SET_ROUND,
-  payload: round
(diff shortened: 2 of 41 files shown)


arpanlaha:
> ok

ok

arpanlaha:
oops did not mean to close

tko22:
why is it cleaner?

arpanlaha:
> what makes trailing commas cleaner?

So if we go from this:
```js
import { 
  a,
  b
}
```

to this:
```js
import { 
  a,
  b,
  c
}
```

In a diff, both the lines with `b` and `c` will be different since b has a comma added. 
Trailing commas will show `c` as the only changed line which is more accurate to the actual changes.

n3a9:
@tko22 I linked an article in the PR description, where it says "cleaner diffs"

Also,  we had a conversation about it [here](https://github.com/hack4impact-uiuc/h4i-recruitment/pull/252#discussion_r294604007).

tko22:
cool, let's do it. Just make sure whenever we do refactoring or linting changes, the changes we make are worth doing, because history gets removed from that (or it's much harder to see who actually made the functional change)

tko22:
let's make sure the nothing regresses like #302

n3a9:
great idea!

josh-byster:
> cool, let's do it. Just make sure whenever we do refactoring or linting changes, the changes we make are worth doing, because history gets removed from that (or it's much harder to see who actually made the functional change)

+1 on this but trailing commas are not too bad since they'll just be a one-line change on those components rather than, say, changing the number of spaces that the files are indented with.

tko22:
yep, it should be fine since this is a linting problem. But also just found #339 so do be very careful.

PR #304, opened 2019-07-18
Current cycle

alan-fang:
Add a `current` field to `Cycle`, which lets us know which Cycles are current and which ones are dated

### DIFF
--- a/backend/src/api/cycle.js
+++ b/backend/src/api/cycle.js
@@ -4,6 +4,8 @@ const { directorsOnly, errorWrap } = require('../middleware')
 const router = express.Router()
 
 // Get all cycles (previous and current)
+// TODO: when authentication server is integrated, only show the cycles that
+// belong to the caller of this endpoint.
 router.get(
   '/',
   [directorsOnly],
@@ -18,6 +20,8 @@ router.get(
 )
 
 // Get a cycle by ID
+// TODO: when authentication server is integrated, only show the cycle id if it belongs
+// to the workspace owned by the caller. Otherwise, 404.
 router.get(
   '/:cycle_id',
   [directorsOnly],
@@ -27,6 +31,33 @@ router.get(
     res.json({
       code: 200,
       result: cycle,
+      success: true
+    })
+  })
+)
+
+// get all cycles belonging to a workspace (either current or current + outdated)
+// TODO: when authentication server is integrated, only show the cycles
+// to the workspace's owner. Otherwise, 404.
+router.get(
+  '/workspace/:workspaceName',
+  [directorsOnly],
+  errorWrap(async (req, res) => {
+    const workspaceName = req.params.workspaceName
+    const cycles = await Cycle.find({ workspaceName, current: req.body.current })
+
+    if (cycles.length === 0) {
+      res.json({
+        code: 404,
+        result: {},
+        success: false
+      })
+      return
+    }
+
+    res.json({
+      code: 200,
+      result: cycles,
       success: true
     })
   })
@@ -50,15 +81,31 @@ router.post(
       response = 'Invalid workspace'
     }
 
-    // TODO: ensure that given workspaces matches one in the database
-    // created by the director
+    // set the last current cycle to not-current
+    Cycle.findOneAndUpdate(
+      { workspaceName: workspace, current: true },
+      { $set: { current: false } },
+      err => {
+        if (err) {
+          res.json({
+            code: 400,
+            message: err.message,
+            result: {},
+            success: false
+          })
+        }
+      }
+    )
+
+    // TODO: when authentication server is integrated, ensure that given workspaces
+    // matches one in the database created by the director
 
     const cycle = new Cycle({
       term: newTerm,
       workspaceName: workspace
     })
-
     await cycle.save()
+
     code = 200
     res.json({
       code,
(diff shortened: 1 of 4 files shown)


backend/src/api/cycle.js
```diff
@@ -32,6 +36,38 @@ router.get(
...
+// to the workspace's owner. Otherwise, 404.
+router.get(
+  '/workspace/:workspaceName',
+  [directorsOnly],
+  errorWrap(async (req, res) => {
+    const workspaceName = req.params.workspaceName
+    const current = req.body.current
+
+    const cycles =
+      current !== null
```
n3a9: nit: check [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) values with just `if (current)`

backend/src/api/cycle.js
```diff
@@ -32,6 +36,38 @@ router.get(
...
+        ? await Cycle.find({ workspaceName, current })
+        : await Cycle.find({ workspaceName })
+
+    if (cycles.length === 0) {
```
n3a9: nit:

```js
return (cycles.length === 0) 
    ? res.json({
           code: 404,
           result: {},
           success: false
       }) 
    : res.json({
         code: 200,
         result: cycles,
         success: true
       })
```

backend/src/api/cycle.js
```diff
@@ -32,6 +36,37 @@ router.get(
...
+        code: 400,
+        message: 'malformed request',
+        success: false
+      })
+    }
+
+    const cycles =
+      current !== null
+        ? await Cycle.find({ workspaceName, current })
+        : await Cycle.find({ workspaceName })
```
chyku: super nitpicky but when reading this it ends up being a double negative for the `else` part

backend/src/api/cycle.js
```diff
@@ -32,6 +36,37 @@ router.get(
   })
 )
 
+// get all cycles belonging to a workspace (either current, outdated, or both)
+// TODO: when authentication server is integrated, only show the cycles
+// to the workspace's owner. Otherwise, 404.
```
chyku: if they request a workspace that doesn't belong to them, should it be a 404 or just an empty response?

backend/src/models/cycle.js
```diff
@@ -3,13 +3,15 @@ const mongoose = require('mongoose')
 /**
  * Holds information about a recruitment cycle; attributes are subject to change.
  *
- * term: (for example) SP19, FA20
+ * term: first 2 letters of the season followed by last 2 digits of the year (i.e. SP19, FA20)
```
chyku: thanks for clarifying the description!

backend/src/api/cycle.js
```diff
@@ -32,6 +36,37 @@ router.get(
...
+  errorWrap(async (req, res) => {
+    const workspaceName = req.params.workspaceName
+    const current = req.body.current
+
+    if (!workspaceName) {
+      res.json({
+        code: 400,
+        message: 'malformed request',
+        success: false
+      })
```
arpanlaha: could be an inline if like what neeraj's PR did

backend/src/api/cycle.js
```diff
@@ -50,15 +85,31 @@ router.post(
...
+      { workspaceName: workspace, current: true },
+      { $set: { current: false } },
+      err => {
+        if (err) {
+          res.json({
+            code: 400,
+            message: err.message,
+            result: {},
+            success: false
+          })
```
arpanlaha: same as above

backend/src/api/cycle.js
angadgarg25: I thought you needed to have an empty `return` statement here after the `res.json` so it doesn't get called twice? 

Also not sure if I agree with inline if - I feel like it makes sense within the `render` function of React, but I think it would make the logic harder to follow in most other places. I'm open to discussion on this point though.

alan-fang: one thing I'm wondering now is why we don't just do 
```
return res.json({...})
```
and instead change `code` and `message`, which I kind of hate

backend/src/api/cycle.js
alan-fang: hmmmm empty response suggests workspace exists but cycles not created, so I think 404 is better

backend/src/api/cycle.js
```diff
@@ -4,6 +4,8 @@ const { directorsOnly, errorWrap } = require('../middleware')
 const router = express.Router()
 
 // Get all cycles (previous and current)
+// TODO: when authentication server is integrated, only show the cycles that
```
tko22: let's link ticket numbers to these TODO items!

backend/src/api/cycle.js
chyku: that's true, what about 401? saying that it's not found is leaving a bad taste in my mouth somehow

alan-fang: I think 403 is actually the one we should use lol

https://stackoverflow.com/questions/50143518/401-unauthorized-vs-403-forbidden-which-is-the-right-status-code-for-when-the-u?rq=1

backend/src/api/cycle.js
alan-fang: ya np

alan-fang: https://github.com/hack4impact-uiuc/h4i-recruitment/issues/305

backend/src/api/cycle.js
chyku: lmao there we have it

Issue #307, opened 2019-07-21
Event Modal Component


alicesf2:
Create reusable modal component for creating new events and event check-in (and other pages that might use modals)

PR #308, opened 2019-07-21
Create separate modal component for use on event pages

alicesf2:
Refs #307

### DIFF
--- a/frontend/src/components/eventsModal.js
+++ b/frontend/src/components/eventsModal.js
@@ -51,7 +51,6 @@ class EventsModal extends Component {
   render() {
     const alert = <Alert color="danger">{this.state.alert}</Alert>
     return (
-      <>
         <Modal isOpen={this.props.isOpen}>
           <ModalHeader>{this.props.title}</ModalHeader>
           <ModalBody>
@@ -80,7 +79,6 @@ class EventsModal extends Component {
             </Button>
           </ModalFooter>
         </Modal>
-      </>
     )
   }
 }
--- a/frontend/src/components/eventsModal.js
+++ b/frontend/src/components/eventsModal.js
@@ -51,34 +51,34 @@ class EventsModal extends Component {
   render() {
     const alert = <Alert color="danger">{this.state.alert}</Alert>
     return (
-        <Modal isOpen={this.props.isOpen}>
-          <ModalHeader>{this.props.title}</ModalHeader>
-          <ModalBody>
-            {this.state.alert && alert}
-            <Form>
-              {this.props.formFields.map(field => (
-                <FormGroup>
-                  <Label>{field.label}</Label>
-                  <Input
-                    name={field.name}
-                    type={field.type}
-                    placeholder={field.placeholder}
-                    onChange={this.props.handleChange}
-                    onKeyPress={this.handleKeyPress}
-                  />
-                </FormGroup>
-              ))}
-            </Form>
-          </ModalBody>
-          <ModalFooter>
-            <Button color="secondary" onClick={this.closeModal}>
-              Cancel
-            </Button>
-            <Button color="primary" onClick={this.handleSubmit}>
-              Submit
-            </Button>
-          </ModalFooter>
-        </Modal>
+      <Modal isOpen={this.props.isOpen}>
+        <ModalHeader>{this.props.title}</ModalHeader>
+        <ModalBody>
+          {this.state.alert && alert}
+          <Form>
+            {this.props.formFields.map(field => (
+              <FormGroup>
+                <Label>{field.label}</Label>
+                <Input
+                  name={field.name}
+                  type={field.type}
+                  placeholder={field.placeholder}
+                  onChange={this.props.handleChange}
+                  onKeyPress={this.handleKeyPress}
+                />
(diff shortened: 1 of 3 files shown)


frontend/src/components/eventsModal.js
```diff
@@ -0,0 +1,94 @@
...
+      name: this.state.name ? this.state.name : '',
+      date: this.state.date ? this.state.date : '',
+      startTime: this.state.startTime ? this.state.startTime : '',
+      endTime: this.state.endTime ? this.state.endTime : '',
+      location: this.state.location ? this.state.location : '',
+      description: this.state.description ? this.state.description : '',
+      fbLink: this.state.fbLink ? this.state.fbLink : ''
+    }
+    const { code } = await createEvent(event)
+    if (code === 500) {
```
angadgarg25: check for any failing request, not just 500. (400 and above basically)

frontend/src/components/eventsModal.js
```diff
@@ -0,0 +1,94 @@
...
+  constructor(props) {
+    super(props)
+    this.state = {
+      alert: ''
+    }
+  }
+
+  addEvent = async () => {
+    const event = {
+      name: this.state.name ? this.state.name : '',
```
alan-fang: replace `this.state.name ? this.state.name : ''` with `this.state.name`, server should catch this error and throw a response

frontend/src/pages/eventOverview.js
```diff
@@ -92,87 +93,56 @@ class EventOverview extends Component {
...
+                  type: 'textarea',
+                  placeholder: 'i.e. Teams present their final product.'
+                },
+                {
+                  label: 'Facebook Event Link',
+                  name: 'fbLink',
+                  type: 'url',
+                  placeholder: 'i.e. https://www.facebook.com/events/2405129342892922/'
+                }
+              ]}
```
angadgarg25: I like this, but I would move this to a different file (in utils or something), and then import it and pass the the constant variable set to this JSON array.

frontend/src/components/eventsModal.js
alan-fang: use the `success` field of the response not the code (also 500 should not map to a user's error). This means that you should be checking fields in backend too!

frontend/src/components/eventsModal.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  render() {
+    const alert = this.state.alert ? <Alert color="danger">{this.state.alert}</Alert> : null
+    return (
+      <>
+        <Modal isOpen={this.props.isOpen}>
+          <ModalHeader>{this.props.title}</ModalHeader>
+          <ModalBody>
+            {alert}
```
alan-fang: get rid of the ternary on 58 and do something like `this.state.alert && {alert}` maybe? seems cleaner to me but is kind of a nit

frontend/src/components/eventsModal.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  handleKeyPress = async event => {
+    if (event.key === 'Enter') {
+      await this.addEvent()
+    }
+  }
+
+  render() {
+    const alert = this.state.alert ? <Alert color="danger">{this.state.alert}</Alert> : null
```
alan-fang: so like this could just be `const alert = <Alert color="danger">{this.state.alert}</Alert>`

frontend/src/pages/eventOverview.js
```diff
@@ -92,87 +93,56 @@ class EventOverview extends Component {
...
-              </ModalFooter>
-            </Modal>
+            <EventsModal
+              title="Add New Event"
+              isOpen={this.state.modal}
+              formFields={[
+                {
+                  label: 'Name',
+                  name: 'name',
+                  type: '',
```
alan-fang: get rid of any fields that are just blank strings

frontend/src/pages/eventOverview.js
```diff
@@ -74,105 +48,27 @@ class EventOverview extends Component {
...
     }
-    const { code } = await createEvent(event)
-    if (code === 500) {
-      this.setState({
-        alert: 'All fields are required.'
-      })
-    } else {
-      Router.push({ pathname: '/eventOverview' })
-    }
+    const { success } = await createEvent(event)
```
angadgarg25: nice use of passing functions as props, this makes the modal even more extendable in the future if we want to use it for other things.

frontend/src/components/eventsModal.js
```diff
@@ -0,0 +1,88 @@
...
+        alert: ''
+      })
+    }
+    this.props.toggle()
+  }
+
+  render() {
+    const alert = <Alert color="danger">{this.state.alert}</Alert>
+    return (
+      <>
```
angadgarg25: The fragment is unnecessary here. The main use case is when you have multiple components at the top most level (like `Head`, `Nav`, etc). Here the only top level component is `Modal` so the fragment is unnecessary.

frontend/src/pages/eventOverview.js
alicesf2: thanks!

PR #311, opened 2019-07-23
Registration + Login Workflows w/ PI Server and Google Auth 

### LINKED ISSUE #263: [Auth Infra] Create Admin Page
Create new React component that admins/co-directors will use to distribute roles to members. This component includes a table that has the members names and represents their year and role in dropdown menus. This component should also include onClick/onChange methods for dropdowns and buttons so that backend integration can be seamless. 

Expected UI can be found [here](https://whimsical.com/HUwX3SrjztKMV8TgQfp9R8).

### LINKED ISSUE #312: Register and Login Workflows with PI Server
Use PI's authentication server to create new login and registration workflows for new members so they can create their own account via either username and password, or their google account.

n3a9:
Closes #312, closes #263 

**Description:** 
This PR includes the following changes:  
- New register/login workflows: users will now be able to create their own accounts by using a username or password, or by connecting their google accounts. Once their account is created, their default role is a member and remain a member until an admin changes it. 
- API calls to deployed backend server: our auth server is now deployed and we can access it via the link in utils/api.js. 
- Wireframe for new admin roles component -> no **real** functionality yet, but coming once implement the backend for new account creation. This will be the data for the adminRoles components 

**Notes:** 
- This PR does not include all of the authentication changes we need, so for now the 'login' tab will remain in the nav bar where you can still log in with your key in case a bug happens. This will eventually change to the same format as the login page 
- We need to create a new table that records the accounts being created. The PI server uses tokens to keep track of each newly created account, so we should record those in order to get the role of each profile in our database. Before this happens, we will be keeping the default sign-in role as just member. 
- Styling is later priority lol 

**Screenshots:** 
Screen Shot 2019-07-23 at 10 51 11 PM
Screen Shot 2019-07-23 at 10 51 19 PM


alan-fang:
Can we get some screenshots? Seems like frontend code was added, would be nice to see the visual changes

caiyingshen:
oops forgot to include, updated

frontend/src/components/nav.js
```diff
@@ -73,6 +74,7 @@ class NavigationBar extends Component {
       Router.push('/dashboard')
 
       this.setState({
+        isDirector: result.is_director,
```
alan-fang: should we still be using `validateKey` if we have the server integrated? Just wondering

frontend/src/pages/index.js
```diff
@@ -1,49 +1,82 @@
...
+      this.setState({ errorMessage: resp.message, showInvalidRequestModal: true })
+      console.log(resp.message)
     } else {
-      this.setState({
-        loading: false
-      })
-      alert("Couldn't log in. Is your key correct?")
+      this.setCookie('token', e.tokenId)
+      this.setCookie('google', true)
+      localStorage.setItem('interviewerKey', 'ohno') // TODO: Create switch statements for roles
```
alan-fang: another note: is localStorage safe? Seems hacky to me and pulled up this article that has some concerns about it, https://dev.to/rdegges/please-stop-using-local-storage-1i04

Thoughts?

frontend/src/utils/api.js
```diff
@@ -9,6 +9,8 @@ const API_URL =
     : 'http://localhost:8080' // make sure your backend is running on this port.
 // if your frontend can't connect, try the normal IP
 
+const AUTH_API_URL = 'https://h4i-infra-server.n3a9.now.sh'
```
angadgarg25: This shouldn't be in the code base, maybe `.env` or something.

frontend/src/pages/index.js
angadgarg25: Also when putting `TODOs` in comments, please create GitHub issues and link them. Even for things you might not get to its very helpful for future development.

frontend/src/pages/index.js
```diff
@@ -52,30 +87,67 @@ class LoginPage extends Component {
...
+                  />
+                </FormGroup>
+                <Button color="outline-secondary" onClick={this.handleSubmit}>
+                  Submit
+                </Button>
+              </Form>
+
+              <GoogleLogin
+                className="btn sign-in-btn"
+                clientId="992779657352-2te3be0na925rtkt8kt8vc1f8tiph5oh.apps.googleusercontent.com"
```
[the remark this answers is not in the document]
angadgarg25: Should this be in version control at all? Jw. It seems like it could be sort of sensitive.

frontend/src/pages/adminRoles.js
```diff
@@ -0,0 +1,78 @@
...
+              {this.state.isEditing && (
+                <td>
+                  <Button>Submit</Button>
+                </td>
+              )}
+            </tr>
+          </tbody>
+        </Table>
+        {this.editButton()}
+        {this.saveButton()}
```
angadgarg25: Whats the reason you aren't rendering these inline with everything else? Also if you have a good reason for this approach these can just be constants rather than function calls probably? Correct me if I'm wrong.

frontend/src/components/nav.js
```diff
@@ -186,6 +188,13 @@ class NavigationBar extends Component {
                       <a className="nav-bar-link pl-3">Rounds</a>
                     </Link>
                   </NavItem>
+                  {(this.state.isDirector || this.state.isLead) && (
```
angadgarg25: Only directors should be able to do this not leads.

frontend/src/pages/adminRoles.js
```diff
@@ -0,0 +1,78 @@
+import React from 'react'
+import { Button, Table } from 'reactstrap'
+import ChangeRole from '../components/changeRole'
+import ChangeYear from '../components/changeYear'
+import Head from '../components/head'
+
+import Nav from '../components/nav'
+
+class AdminRoles extends React.Component {
```
angadgarg25: What if I were to manually type adminRoles into the path? Doesn't seem like anything is stopping me from accessing the page as a member.

frontend/src/components/nav.js
```diff
@@ -40,6 +40,7 @@ class NavigationBar extends Component {
   constructor(props) {
     super(props)
     this.state = {
+      isDirector: false,
```
angadgarg25: Can we use the Enum here too? Otherwise we could run into weird undefined states since there are 4 combinations of these 2 boolean values, but only 3 roles. Even if that doesn't impact the logic right now, it could impact future logic, and an enum here would be cleaner.

angadgarg25:
Also the `login` Nav component currently still asks for your key.

frontend/src/utils/api.js
angadgarg25: Maybe this is fine, depends on what someone could do with the link.

frontend/src/pages/index.js
```diff
@@ -52,30 +85,67 @@ class LoginPage extends Component {
...
+          </Card>
+          <Button color="outline-primary" onClick={() => Router.push('/register')}>
+            {"Don't have an account? Register here!"}
+          </Button>
+          <Modal autoFocus={false} isOpen={this.state.showInvalidRequestModal}>
+            <ModalHeader>{'There was an error in your request.'}</ModalHeader>
+            <ModalBody>
+              {this.state.errorMessage
+                ? this.state.errorMessage
+                : 'There was an error in your request'}
```
arpanlaha: stylistic change, could be the following: 
```js
{this.state.errorMessage || 'There was an error in your request'}
```
Is that something we'd want or should we stick to ternary?

frontend/src/pages/register.js
```diff
@@ -0,0 +1,211 @@
...
+            <ModalFooter>
+              <Button onClick={this.handleInvalidPasswordModalClose}>Close</Button>
+            </ModalFooter>
+          </Modal>
+          <Modal autoFocus={false} isOpen={this.state.showInvalidGoogleRequestModal}>
+            <ModalHeader>{'There was an error in your request.'}</ModalHeader>
+            <ModalBody>
+              {this.state.errorMessage !== null
+                ? this.state.errorMessage
+                : 'There was an error in your request'}
```
arpanlaha: same as above

frontend/src/pages/index.js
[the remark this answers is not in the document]
tko22: this is fine. we shouldn't expose client secret

frontend/src/utils/enums.js
```diff
@@ -52,6 +52,12 @@ export const rolesEnum = {
   UI: 'UI/UX Designer'
 }
 
+export const permissionRolesEnum = {
+  M: 'Member',
+  L: 'Lead',
+  D: 'Director'
```
tko22: nit. i’d rather have it be `MEMBER` `LEAD` etc instead of M,L,D because it’s more clear what those enums mean when you are reading code. For instance, permissionRolesEnum.MEMBER is easier to understand than permissionRolesEnum.M. When reading the code i would need to go to the enums page to understand what `M` means

frontend/src/utils/api.js
```diff
@@ -250,6 +252,55 @@ function deleteReferral(candidateID: string) {
   }).then(res => res.json())
 }
 
+function registerUser(email: string, password: string, role: string) {
+  console.log(`Creating new user: ${email}`)
```
tko22: i like the logs

frontend/src/pages/register.js
```diff
@@ -0,0 +1,213 @@
+// adapted heavily from H4I PI's register workflow examples found here:
```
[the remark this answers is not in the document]
caiyingshen: I thought about it, but personally it makes more sense to have 2 straightforward components than 1 component that has a lot of if's and ternaries, which is what I saw this turning into. I also think if we want to add more functionality to one of these pages but not the other (which I can see happening for the register page), its more extendable this way

frontend/src/pages/index.js
```diff
@@ -1,49 +1,84 @@
...
+      this.setState({ errorMessage: resp.message, showInvalidRequestModal: true })
+      console.log(resp.message)
     } else {
-      this.setState({
-        loading: false
-      })
-      alert("Couldn't log in. Is your key correct?")
+      this.setCookie('token', e.tokenId)
+      this.setCookie('google', true)
+      localStorage.setItem('interviewerKey', 'ohno') // TODO: Create switch statements for roles
```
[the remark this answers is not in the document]
caiyingshen: actually, I'm going to try to make the workflow work without this line but might involve a lot of other changes

frontend/src/pages/register.js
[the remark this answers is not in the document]
alan-fang: Sounds good, you made a lot of valid points here

frontend/src/pages/index.js
n3a9: Much prefer this way as well

frontend/src/components/nav.js
caiyingshen: If it's okay with you, I'd like to make this change in a separate PR. I'm trying to add the role variable into the redux state instead of having booleans here and in other components, but it's making this PR a lot larger than it should be.

frontend/src/utils/api.js
caiyingshen: I'm not sure if this counts as an environment variable. I'm pretty sure it has the same amount of functionality as the other API URL in this file, but I can move both of them to the .env file if you want

frontend/src/components/nav.js
alan-fang: ok sounds good

what do you mean by `until the backend is implemented` though? What is there left to do? Just want to keep an eye on the blockers for me and @chyku

frontend/src/pages/index.js
n3a9: I think that this is a larger refactor that can be considered in another PR.

frontend/src/pages/adminRoles.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react'
+import { Button, Table } from 'reactstrap'
+import ChangeRole from '../components/changeRole'
+import ChangeYear from '../components/changeYear'
+import Head from '../components/head'
+
```
alan-fang: why the blank line?

frontend/src/pages/adminRoles.js
```diff
@@ -0,0 +1,67 @@
...
+class AdminRoles extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = {
+      isEditing: false
+    }
+  }
+
+  render() {
+    const sampleMemberId = 1234
```
alan-fang: nit: move this to the top of the file and make it all caps `SAMPLE_MEMBER_ID`

frontend/src/pages/index.js
```diff
@@ -1,49 +1,89 @@
...
+      cookie.set(key, value, {
+        expires: cookieExpirationDays,
+        path: '/'
       })
-      Router.push('/dashboard')
+    }
+  }
+
+  handleGoogle = async e => {
+    const memberKey = 'ohno'
```
alan-fang: probably also want to move this to the top of the file and make it `MEMBER_KEY`

frontend/src/pages/adminRoles.js
n3a9: Please preface with `nit`.

frontend/src/pages/index.js
```diff
@@ -1,49 +1,89 @@
...
+    const result = await loginGoogleUser(e.tokenId)
+    const resp = await result.json()
```
alan-fang: name this `response` instead, `resp` could confuse someone

frontend/src/pages/index.js
```diff
@@ -1,49 +1,89 @@
...
+    this.setState({ [name]: value })
   }
 
-  // handles when user presses "Enter" when input is focused
-  _handleKeyPress = e => {
-    if (e.key === 'Enter') {
-      this.handleSubmit()
-    }
+  handleSubmit = () => {
+    const memberKey = 'ohno'
```
alan-fang: so if you move this to the top as `MEMBER_KEY` you won't have to redefine it in both `handleGoogle` and `handleSubmit`

frontend/src/pages/index.js
```diff
@@ -1,49 +1,89 @@
...
+    const { email, password } = this.state
+    console.log(`Logging in ${email}`)
+    loginUser(email, password).then(resp => {
+      if (!resp.success) {
```
alan-fang: `response`

frontend/src/pages/index.js
```diff
@@ -52,30 +92,62 @@ class LoginPage extends Component {
...
-              <h4>Enter Key:</h4>
-              <Input
-                autoFocus={true}
-                type="text"
-                value={this.state.currentKey}
-                onChange={this.onTextChange}
-                name="Input Key"
-                placeholder="Input Your Key"
-                onKeyPress={this._handleKeyPress}
+          <Card className="login-card" style={{ width: '400px', height: '60%' }}>
```
alan-fang: can you move this to the css file?

frontend/src/pages/index.js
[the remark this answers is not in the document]
alan-fang: would still greatly prefer this to be put in a constants section, or a config or something!

frontend/src/pages/register.js
```diff
@@ -0,0 +1,212 @@
...
+  Card,
+  CardBody,
+  CardTitle,
+  Modal,
+  ModalBody,
+  ModalFooter,
+  ModalHeader
+} from 'reactstrap'
+
+const EMAIL_REGEX =
```
alan-fang: wow love this

frontend/src/pages/register.js
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  handleChange = event => {
+    const value = event.target.value
+    const name = event.target.name
+    this.setState({ [name]: value })
+  }
+
+  handleGoogle = async e => {
+    const memberKey = 'ohno'
```
alan-fang: re: my comment about constants

frontend/src/pages/register.js
```diff
@@ -0,0 +1,212 @@
...
+      // set google to true so server knows to send the request to google
+      this.setCookie('google', true)
+      // set localStorage value so it's valid across the whole site
+      localStorage.setItem('interviewerKey', memberKey) // TODO: Create switch statements for roles - Issue #314
+      Router.push('/dashboard')
+    }
+  }
+
+  handleSubmit = () => {
+    const memberKey = 'ohno'
```
alan-fang: re: my comment about constants

frontend/src/pages/register.js
```diff
@@ -0,0 +1,212 @@
...
+
+class RegisterPage extends Component {
+  constructor(props) {
+    super(props)
+    this.state = {
+      email: '',
+      password: '',
+      passwordVerification: '',
+      errorMessage: '',
+      showInvalidPasswordModal: false,
```
alan-fang: would probably be better to have 1 variable displaying which type of modal to show: `InvalidPasswordModal`, `InvalidRequestModal`, `InvalidGoogleRequestModal`, or `None`. Unless it's possible to have multiple modals?

frontend/src/pages/register.js
```diff
@@ -0,0 +1,212 @@
...
+    this.setState({ showInvalidGoogleRequestModal: false })
+  }
+
+  render() {
+    return (
+      <>
+        <Head />
+        <Nav />
+        <Container>
+          <Card className="interview-card" style={{ width: '400px', height: '60%' }}>
```
alan-fang: re: my comment about avoiding inline styling

frontend/src/pages/register.js
```diff
@@ -0,0 +1,212 @@
...
+                  size="lg"
+                  onClick={this.handleSubmit}
+                  style={{ float: 'left', width: '48%' }}
+                >
+                  Register
+                </Button>
+              </Form>
+              <GoogleLogin
+                className="btn-lg sign-in-btn"
+                clientId="992779657352-2te3be0na925rtkt8kt8vc1f8tiph5oh.apps.googleusercontent.com"
```
alan-fang: re: my comment about storing this kind of info in a config

frontend/src/utils/api.js
```diff
@@ -250,6 +252,55 @@ function deleteReferral(candidateID: string) {
...
+      headers: {
+        'content-type': 'application/json'
+      },
+      body: JSON.stringify({
+        tokenId: tokenId,
+        role: 'member'
+      })
+    })
+  } catch (err) {
+    console.log(err)
```
alan-fang: just as a general comment, is there a reason why we log the error here and not in `loginUser`, etc? Also this seems like we silently fail on the frontend — that's definitely not good UX for a user

frontend/src/components/changeRole.js
```diff
@@ -0,0 +1,17 @@
+// Change Role dropdown
+// handler function must be included for logic
+import React from 'react'
+import { permissionRolesEnum } from '../utils/enums'
+
+const ChangeRole = ({ memberID, role, handleChange }) => (
```
alan-fang: any reason why we can't include the handler function here?

frontend/src/components/changeYear.js
```diff
@@ -0,0 +1,18 @@
+// Change Year dropdown
+// handler function must be included for logic
+import React from 'react'
+import { yearsEnum } from '../utils/enums'
+
```
alan-fang: `:same_here:`

frontend/src/pages/adminRoles.js
alan-fang: (you can handle this in a follow-up PR so we can ship this one faster)

frontend/src/pages/index.js
alan-fang: (you can handle this in a follow-up PR so we can ship this one faster)

frontend/src/pages/index.js
alan-fang: (you can handle this in a follow-up PR so we can ship this one faster)

frontend/src/pages/index.js
alan-fang: (you can handle this in a follow-up PR so we can ship this one faster)

frontend/src/pages/index.js
alan-fang: (you can handle this in a follow-up PR so we can ship this one faster)

frontend/src/pages/index.js
alan-fang: (you can handle this in a follow-up PR so we can ship this one faster)

frontend/src/pages/register.js
alan-fang: (you can handle this in a follow-up PR so we can ship this one faster)

frontend/src/pages/register.js
alan-fang: (you can handle this in a follow-up PR so we can ship this one faster)

frontend/src/pages/register.js
alan-fang: (you can handle this in a follow-up PR so we can ship this one faster)

frontend/src/pages/register.js
alan-fang: (you can handle this in a follow-up PR so we can ship this one faster)

frontend/src/pages/register.js
alan-fang: (you can handle this in a follow-up PR so we can ship this one faster)

frontend/src/components/changeRole.js
caiyingshen: i think it's very possible for us to reuse this component in the future and we can't guarantee that it'll have the same handler function

frontend/src/pages/register.js
caiyingshen: #326

frontend/src/components/nav.js
caiyingshen: The PI's server requires a tokenId in order to retrieve a profile, so we need to create a table to hold the tokens for the users so we can get each user and their role. The backend is basically making the `get` endpoint for the tokens

frontend/src/utils/api.js
angadgarg25: Nvm, I think this is fine for now. I just need a quick readme on the setup process for auth for this project specifically, and then it should be good to go from my end. I'll just setup the auth server from my end and switch in the URL.

angadgarg25:
Actually I've set this up before, I just need to know what roles you've used (names). I would assume its safe to use 4 (Director, Lead, Member, Pending), let me know if that works for you guys,

caiyingshen:
@angadgarg25 I'll add a blurb to the README, but @n3a9 is the one who set up the backend server so I'm not 100% sure on some of the details. I think it's easy enough that if we want to change our roles we can set up a new server or something so that should be fine. I'll add 'PENDING" to the enum, but for clarification what rights should someone with a pending role have? Should they just see an empty/waiting page until an admin assigns them a role?

angadgarg25:
@caiyingshen Yeah they basically should actually be able to see any info pretty much. I thin currently thats what our member permissions are too, but eventually we might change member permissions a little, while keeping pending members with no permissions.

frontend/src/pages/index.js
```diff
@@ -1,49 +1,89 @@
...
-      Router.push('/dashboard')
+    }
+  }
+
+  handleGoogle = async e => {
+    const result = await loginGoogleUser(e.tokenId)
+    const response = await result.json()
+    if (!response.success) {
+      this.setState({ errorMessage: response.message, showInvalidRequestModal: true })
+      console.log(response.message)
```
tko22: would recommend you remove this logging statement but issok

caiyingshen:
@angadgarg25 ah okay sounds good, i'm going to scope that in with #314 so this PR doesn't continue ballooning

frontend/src/utils/api.js
angadgarg25: Change to `'h4i-portal-infra-server.now.sh'`

frontend/src/pages/register.js
angadgarg25: Change to `'409847273934-jmhjkeu77d3cqr32sh3vpl3ogh2f4dev.apps.googleusercontent.com'`

PR #315, opened 2019-07-28
Automatic Interview Scheduling

### LINKED ISSUE #250: Scheduling Part 2
Sub-epic of #171 and continuation of #202. We have part of this finished, given an interview schedule on excel, we can add it to the UI. However, we’d like to take availabilities of interviewers and interviewees in the spreadsheet format output by Doodle, and generate a interview schedule that we directly add to the UI. Additionally, right now we do not connect either the interviewers or interviewees from the schedule to their user/candidate model (table), so we’d like to do this as a way to have the schedule-viewing UI show a hack4impact member only the interviews they have, and link directly to allow them to fill out an interview form/notes for that candidate.

Things to consider: We have one doodle for interviewers, and one for interviewees. Interviewees only get 1 slot for a specific round. Interviewers may have multiple, but should try to be relatively  equal if possible. (not 1 person doing 8 interviews, etc.). Additionally, we have the mass interview round, where a slot will have multiple interviewers and interviewees. Basically, number of interviewers and interviewees for a slot should be parameters. Finally once output into a schedule, directors should be able to modify things manually, so that in case an interviewer or interviewee needs to switch, they can just fix those without affecting the whole schedule. Currently schedules are stored in the model: [`futureInterview.js`](https://github.com/hack4impact-uiuc/h4i-recruitment/blob/master/backend/src/models/futureInterview.js).

To generate the schedule let's take a note from 374. Don't create your own algorithm from scratch. Best case find an algorithm online that we can use (reference it in comments maybe?), otherwise find an algorithm that does something very similar and modify slightly to have it work. Note that modifying could be modifying the algorithm or modifying the input (and the ladder might be the easier way to go about it since you'll almost certainly have to do this regardless).

Once the schedule is created it should be displayable on the frontend. This is the [current display](https://github.com/hack4impact-uiuc/h4i-recruitment/blob/master/frontend/src/pages/interviewschedule.js) for the schedule, but we can update this to link directly to the candidate, and allow a user to filter for their own interviews.

Being able to press a button to email everyone with their times would be a nice to have feature, but not as high priority,

josh-byster:
Resolves #250

Screen Shot 2019-08-29 at 1 12 31 AM
Screen Shot 2019-08-29 at 1 12 24 AM


backend/src/api/futureInterview.js
```diff
@@ -93,35 +45,182 @@ router.post(
...
+        code: 400,
+        message: 'Error.',
+        result: {},
+        success: false
+      })
+    }
+  })
+)
+
+let uploadInterviewAvailability = async (obj, isInterviewerList) => {
```
tko22: make this a constant?

backend/src/api/futureInterview.js
```diff
@@ -93,35 +45,182 @@ router.post(
...
+  let availability = new InterviewAvailability({
+    type: isInterviewerList ? 'INTERVIEWERS' : 'CANDIDATES',
```
tko22: I would put the strings `INTERVIEWERS` and `CANDIDATES` in constants

backend/src/api/futureInterview.js
```diff
@@ -93,35 +45,182 @@ router.post(
...
-  [leadsOnly],
+router.post(
+  '/uploadCandidates',
   errorWrap(async (req, res) => {
-    const futureInterviews = await FutureInterview.find()
-    let interviews = []
-    let code = 500
-    // TODO: do stuff
-    // retrieve list of FutureInterviews and return them
+    var workbook = XLSX.read(req.body['data'], { type: 'binary' })
```
n3a9: Prefer `let` over `var` for [tighter scoping](https://www.geeksforgeeks.org/difference-between-var-and-let-in-javascript/)

n3a9: But I also don't see these modified after creation - is there a reason why we're not using `const`?

frontend/src/pages/interviewschedule.js
```diff
@@ -119,16 +141,48 @@ class InterviewSchedule extends Component {
         <hr />
         <Container>
           <form onSubmit={this.uploadSchedule}>
-            <>
+            <div>
```
n3a9: Prefer `<>` over `<div>`

backend/src/api/futureInterview.js
josh-byster: Yep - that's definitely a typo. Good catch.

frontend/src/utils/api.js
```diff
@@ -40,30 +40,58 @@ function getEventAttendees(id: string) {
...
+  }).then(res => res.json())
+}
+
+const readUploadedFile = inputFile => {
+  const reader = new FileReader()
+
+  return new Promise((resolve, reject) => {
+    reader.onerror = () => {
+      reader.abort()
+      reject(new DOMException('Problem parsing input file.'))
```
josh-byster: TODO: Change from `DOMException` to something more closely matching the error happening.

backend/src/utils/schedule/generate_schedule.py
```diff
@@ -0,0 +1,55 @@
+from pyschedule import Scenario, solvers, plotters, alt
```
josh-byster: @achcello - how should we have this file version controlled since its deployment will be controlled via AWS?

frontend/src/pages/interviewschedule.js
tko22: i would mention that it's `nit` for these, especially when you would preface it with "Prefer"

backend/src/utils/schedule/generate_schedule.py
angadgarg25: We have an H4I AWS, I can move it over there. I'd add to one of the README's explaining setup for that.

backend/src/api/futureInterview.js
```diff
@@ -1,71 +1,28 @@
...
+const fetch = require('isomorphic-unfetch')
+const moment = require('moment')
 const csv = require('csv-array')
 const { errorWrap, leadsOnly } = require('../middleware')
-const { FutureInterview } = require('../models')
+const { FutureInterview, InterviewAvailability } = require('../models')
 const keyPath =
   process.env.NODE_ENV === 'test' ? '../../tests/artifacts/test-keys.json' : process.env.KEY_JSON
+const schedulerUrl = process.env.SCHEDULER_URL
+const schedulerApiKey = process.env.SCHEDULER_API_KEY
```
angadgarg25: Send me updated `.env` file so I can share and deploy when need be.

backend/src/utils/schedule/generate_schedule.py
angadgarg25: Sorry misunderstood. But yes, if you could write a README for the setup details we can move over to that ASAP, not a blocker though. In terms of placement for this, I don't think it makes sense to keep it in here. I'd actually put it in another top-level directory in this repo (e.g. `scheduler`). Then we can also easily add this README there with setup instructions for this part, since it is a separate deployment.

backend/src/utils/schedule/generate_schedule.py
```diff
@@ -0,0 +1,55 @@
...
+final = [[S.resources()[x[0]],S.resources()[x[1]]] for x in name_perms]
+
+for t in task_list:
+    random.shuffle(resource_list)
+    t += alt(resource_list[:10]), alt(resource_list[10:])
+
+
+# compute and print a schedule
+
+solvers.mip.solve(S,random_seed=random.randint(1,1000))
```
alan-fang: Just a small thought, does this handle interviewer preferences? i.e. we don't want people interviewing their friends

frontend/src/pages/interviewschedule.js
```diff
@@ -98,21 +146,30 @@ class InterviewSchedule extends Component {
     }
   }
 
-  async componentDidMount() {
+  populateInterviewSchedules = async () => {
     const res = await getInterviewSchedule()
+    if (res.code !== 200) {
```
alan-fang: don't use `code`, use the `success` field; I know that you only use 200 in your backend but it's better practice to generalize beyond a specific response code

frontend/src/pages/interviewschedule.js
```diff
@@ -121,14 +178,46 @@ class InterviewSchedule extends Component {
           <form onSubmit={this.uploadSchedule}>
             <>
               <h2>Upload a New Schedule</h2>
+              <h4>Candidates</h4>
+              <input
+                ref={ref => {
+                  this.candidateInput = ref
+                }}
+                onChange={_ => this.setState({ candidateFileSelected: true })}
```
alan-fang: out of curiosity, what's the diff between `() =>` and `_ => `?

frontend/src/utils/api.js
```diff
@@ -55,30 +55,58 @@ function getEventAttendees(id: string) {
...
-      },
-      body: JSON.stringify({ schedule: scheduleString })
-    })
-      .then(res => res.json())
-      .then(success => console.log(success))
-      .catch(error => console.log(error))
-  }
-
-  reader.readAsText(file)
+async function addInterviewerSchedules(file: File) {
```
alan-fang: you can probably merge this function with the one below; something like
```
addSchedules(file: File, scheduleType) 
```
and based on the the string passed in fetch to 
```
${API_URL}/schedule/upload${scheduleType}/?key=${getKey()}
```

frontend/src/utils/api.js
```diff
@@ -55,30 +55,58 @@ function getEventAttendees(id: string) {
...
+  }).then(res => res.json())
+}
+
+const readUploadedFile = inputFile => {
+  const reader = new FileReader()
+
+  return new Promise((resolve, reject) => {
+    reader.onerror = () => {
+      reader.abort()
+      reject(new TypeError('Problem parsing input file.'))
```
alan-fang: Maybe there's more detailed error information to pass upstream? This TypeError is not particularly descriptive. Normally I'd just let it go, but I have a sneaky feeling that unexpected issues may come up with uploading and parsing files...

backend/src/utils/schedule/generate_schedule.py
josh-byster: It doesn't currently - good point. Was thinking we could make one-off changes like this manually but maybe it's something we can incorporate. @achcello what are your thoughts?

frontend/src/pages/interviewschedule.js
josh-byster: `()` implies the signature takes in no params, while `_` implies that it takes in some param (ex. `(a) =>`) but to discard that param rather than store it in some variable since it's not used.

frontend/src/utils/api.js
josh-byster: Yeah I definitely agree these can be combined.

frontend/src/utils/api.js
josh-byster: Sure - this will `reject` upon failing to do an asynchronous read of the file passed into the server, rather than an error during the parsing o the file itself, so maybe I can say "Problem reading input file." 
What do you think?

alan-fang: That sounds good to me 👍🏻

backend/src/utils/schedule/generate_schedule.py
alan-fang: There doesn't have to be an action item as a follow-up, just making sure that this is on the radar!

josh-byster:
> Would also be nice if you added backend testing and a screenshot of the frontend changes you made

@alan-fang Yeah I agree, however I'm struggling to find ways to test this given that the entire integration is rather complex. Any ideas on what might be good functionality to have unit tested?

alan-fang:
> > Would also be nice if you added backend testing and a screenshot of the frontend changes you made
> 
> @alan-fang Yeah I agree, however I'm struggling to find ways to test this given that the entire integration with Doodle is rather complex and would be hard to test that uploading. Any ideas on what might be good functionality to have unit tested?

hmmmm looks like we might need to introduce mocking into our API tests for something like this. That's a little complex to do for this specific PR so for now let's just test really simple cases (i.e that GET and POST return 200 when data is supplied, etc). 

I think that `generate_schedule.py` still should have some test cases (and those should be easy to test) though.

Of course, if you're in a big hurry to merge this, I'll approve it, but then I think your testing effort should be in a follow-up PR.

josh-byster:
@alan-fang tests are done! Feel free to merge if you feel it's ready.

alan-fang:
Yeah go ahead and merge, it's already been approved 😃

PR #318, opened 2019-07-28
Front end workspaces

### LINKED ISSUE #280: Create frontend interface for admins to change workspaces

chyku:
Part of #280

It's kind of ugly so if anyone has some UI/UX improvement suggestions hmu! Also I'm sorry this PR turned out to be huge

- [x] Display current workspaces (including API calls to get the cycles, etc.)

- [x] Add ability to create workspaces + cycles in frontend 

Default page: 
image

Create cycle modal:
image

Create workspace modal:
image


alan-fang:
do you mind if I merge a Workspace change in first? I've got one almost ready for review, and then you can rebase and work off of that

chyku:
> do you mind if I merge a Workspace change in first? I've got one almost ready for review, and then you can rebase and work off of that

the pr looks great- it'll be really helpful once it's merged!

backend/src/models/user.js
```diff
@@ -9,6 +9,6 @@ const User = new mongoose.Schema({
   email: String,
   key: String,
   role: String,
-  workspaceId: String
+  cycleId: mongoose.Schema.Types.ObjectId
```
alan-fang: why is user bound to `cycleId` and not `workspaceId`?

frontend/src/pages/workspaces.js
```diff
@@ -0,0 +1,211 @@
...
+      workspaceModal: false,
+      cycleModal: false,
+    }
+  }
+
+  async componentDidMount() {
+    // get workspaces and cycle, if applicable
+    const workspaceRes = await getWorkspaces()
+    if (workspaceRes) {
+      workspaceRes.result.forEach(async workspace => {
```
arpanlaha: why is workspace async? and this could probably be more intuitive as the following:
```js
if (workspaceRes) {
  this.setState({ 
    workspaces: workspaceRes.result.map(workspace => workspace.name),
  });
}
```

backend/src/models/user.js
chyku: oh whoops i went through and changed all of them without realizing what User was- good catch

frontend/src/pages/workspaces.js
chyku: i don't remember why i made it async, but that suggestion seems to work

PR #319, opened 2019-07-29
Workspaces API Layer

### LINKED ISSUE #321: Workspace API Layer
api method calls for workspaces

alan-fang:
Resolves #321
API layer for workspace requests

### DIFF
--- a/backend/src/api/workspaces.js
+++ b/backend/src/api/workspaces.js
@@ -26,17 +26,7 @@ router.get(
   '/:workspaceName',
   [directorsOnly],
   errorWrap(async (req, res) => {
-    const workspaceName = req.params.workspaceName
-
-    if (!workspaceName) {
-      return res.json({
-        code: 400,
-        message: 'malformed request',
-        success: false
-      })
-    }
-
-    const workspace = await Workspace.find({ name: workspaceName })
+    const workspace = await Workspace.find({ name: req.params.workspaceName })
 
     res.json({
       code: 200,
@@ -56,7 +46,7 @@ router.post(
     if (!owner || !workspaceName) {
       return res.json({
         code: 400,
-        message: owner ? 'Invalid owner name' : 'Invalid workspace name',
+        message: owner ? 'Invalid workspace name' : 'Invalid owner name',
         success: false
       })
     }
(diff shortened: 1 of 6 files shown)


backend/src/api/workspaces.js
```diff
@@ -0,0 +1,103 @@
...
+  '/:workspaceName',
+  [directorsOnly],
+  errorWrap(async (req, res) => {
+    const workspaceName = req.params.workspaceName
+
+    if (!workspaceName) {
+      return res.json({
+        code: 400,
+        message: 'malformed request',
+        success: false
```
chyku: I don't think this ever hits because if someone tries to hit the endpoint "/workspace/" it'll direct you to the first endpoint

backend/src/api/workspaces.js
```diff
@@ -0,0 +1,103 @@
...
+router.post(
+  '/',
+  [directorsOnly],
+  errorWrap(async (req, res) => {
+    const owner = req.body.owner
+    const workspaceName = req.body.name
+    if (!owner || !workspaceName) {
+      return res.json({
+        code: 400,
+        message: owner ? 'Invalid owner name' : 'Invalid workspace name',
```
chyku: i think this should be the other way around (if owner exists the msg is 'invalid workspace')

backend/src/api/workspaces.js
```diff
@@ -0,0 +1,103 @@
...
+  })
+)
+
+// Transfer ownership of a workspace to someone else
+router.put(
+  '/transfer/:workspaceName',
+  [directorsOnly],
+  errorWrap(async (req, res) => {
+    const newOwner = req.body.owner
+    const name = req.params.workspaceName
```
chyku: maybe rename `name` bc it can be confusing to have both an owner and a workspace name

backend/src/api/workspaces.js
```diff
@@ -0,0 +1,103 @@
...
+
+    if (!newOwner || !name) {
+      return res.json({
+        code: 400,
+        message: newOwner ? 'Invalid owner name' : 'Invalid workspace name',
```
chyku: same as above

backend/src/api/workspaces.js
```diff
@@ -0,0 +1,103 @@
...
+        success: false
+      })
+    }
+
+    // todo: ensure that new owner is a director & belongs in the same org
+    await Workspace.findOneAndUpdate({ name }, { $set: { owner: newOwner } }, { new: true })
```
chyku: should the new workspace be returned? otherwise would it be necessary to have `{ new : true }`

backend/src/api/workspaces.js
```diff
@@ -0,0 +1,103 @@
...
+
+    res.json({
+      code: 200,
+      success: true
+    })
```
chyku: missing a message and/or result obj in response (#69)

backend/src/api/workspaces.js
```diff
@@ -0,0 +1,103 @@
...
+  })
+)
+
+// Create a new workspace
+router.post(
+  '/',
+  [directorsOnly],
+  errorWrap(async (req, res) => {
+    const owner = req.body.owner
+    const workspaceName = req.body.name
```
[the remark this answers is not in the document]
alan-fang: I think we'll probably use something from @caiyingshen and @n3a9 's work, so I'll take a closer look at their ongoing PR and add the appropriate `TODO`

backend/src/api/workspaces.js
alan-fang: lolol yeah good catch

backend/src/api/workspaces.js
alan-fang: good point, I'm just going to delete `{ new : true }`

backend/src/api/workspaces.js
alan-fang: nice catch!

backend/src/api/workspaces.js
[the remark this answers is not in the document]
angadgarg25: link the issue number as well next to the TODO

PR #324, opened 2019-08-07
Workspace filter for dashboard

### LINKED ISSUE #323: Allow directors to filter candidates by workspace
- Adding to the FilterComponent to allow us to choose which workspace to view candidates from

alan-fang:
Resolves #323

Screen Recording 2019-08-06 at 11 10 21 PM

Added the ability to filter by workspace in the portal dashboard.


frontend/src/components/filterComponent.js
```diff
@@ -106,6 +133,25 @@ class FilterComponent extends Component<Props> {
...
+            <h5>Workspace</h5>
+
+            <p>
+              <select
+                name="workspaces"
+                onChange={this.handleSelectedChange}
+                className="workspace-selector"
+              >
+                <option selected>None</option>
+                {this.state.workspaces.map(obj => (
```
alicesf2: maybe change "obj" to a more specific name like "workspace"

frontend/src/actions/index.js
```diff
@@ -10,6 +10,7 @@ import {
   fetchAllCandidates,
   addFilter,
   removeFilter,
+  resetFilter,
```
arpanlaha: not within the scope of this pr but a general comment - why don't we just `export {...} from "./actionCreators.js"` instead of repeating everything twice?

frontend/src/reducers/initialState.js
```diff
@@ -9,6 +9,7 @@ import {
   selectByEnum,
   compareByEnum
 } from '../utils/enums'
+import { getWorkspaces } from '../utils/api'
```
arpanlaha: unused import?

frontend/src/components/filterComponent.js
```diff
@@ -36,6 +38,18 @@ type Props = {
...
+  constructor(props) {
+    super(props)
+    this.state = {
+      workspaces: []
+    }
+  }
+
+  async componentDidMount() {
+    const workspaces = await getWorkspaces()
+    this.setState({ workspaces: workspaces.result })
```
arpanlaha: will a failed request break `this.state.workspaces.map()`?

frontend/src/actions/index.js
alan-fang: Haven't taken a look into this but I can loop it in to my cleanup commit if it's not too much work

frontend/src/components/filterComponent.js
alan-fang: depends on what you mean by "failed". If you mean that the request returns nothing, then no it doesn't break. If you mean that there's some external error (i.e. network failure), then yes, it will probably break, but this is mostly consistent behavior with other network calls in our web frontend:

https://github.com/hack4impact-uiuc/h4i-recruitment/blob/4878fad7fe1a033f0e48adc42d5cd7e7d2e1a374/frontend/src/pages/stats.js#L27

frontend/src/actions/index.js
alan-fang: it is done

PR #329, opened 2019-08-12
Add ability to run with Docker

### LINKED ISSUE #260: Docker Compose
I think it would be nice if we spun up at least two Docker containers — one for frontend, one for backend, and potentially one for Mongo if we're not connecting to a remote DB for dev, and tie them together with Docker Compose. Ideally, we'd have it so that we could run `docker-compose up` and have everything up and running locally, with consistency amongst all developers.

### LINKED ISSUE #328: API port and misc. debug flags should be environment vars
It would be nice to be able to quickly switch the port using an environment variable, and also turn off and on some misc. debug flags. For example, we may not want to see the debug level for redux state output, so factoring out pieces like that which are currently hardcoded would be ideal.

josh-byster:
Resolves #260

This change allows us to run the frontend and backend all at once with no installation needed, just one command: `docker-compose up`! 🎉
Also we have the uniformity and reproducibility of bugs among all our developers using Docker with this change.

This is meant to be run in a dev environment, so we use volumes to allow changes made locally to be reflected in the container. 

This also addresses factoring out environment variables (fix #328)

Screen Shot 2019-08-12 at 3 46 41 AM

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -2,6 +2,53 @@
 
 ## Setup
 
+In order to have all the secret keys set up properly, you must create a `.env` file and a `keys.json` file to the root of this directory (`./backend`).
+
+Example setup:
+
+In `.env`:
+```
+MONGO_UURL=mongodb://mongouser:mykey@somedb.mlab.com:port/db
+MONGO_DURL=mongodb://mongouser:mykey@somedb.mlab.com:port/db
+MONGO_URL=mongodb://mongouser:mykey@somedb.mlab.com:port/db
+LEAD_SUFFIX=l
+DIRECTOR_SUFFIX=d
+KEY_JSON=../../keys.json
+SCHEDULER_URL=https://scheduler-URI.amazonaws.com/endpoint
+SCHEDULER_API_KEY=abcd1234
+```
+
+In `keys.json`:
+```
+{
+  "keys": [
+    {
+      "name": "Director 1",
+      "key": "1234"
+    }
+  ]
+}
+```
+
+### Docker
+
+To run with [Docker](https://www.docker.com/) (and start the frontend and backend all at once with one command):
+```
+docker-compose up
+```
+
+And when you're done, you can run:
+```
+docker-compose down -v
+```
+
+Note that if there's a change in the `package.json` file (such as a new module added into the repository), instead of running `yarn add` to re-install, just run `docker-compose up --build`. 
+
+
+### Manual
+
+You must install `dotenv-cli`:
+
 ```sh
 yarn global add dotenv-cli
 ```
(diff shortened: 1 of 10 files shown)


backend/README.md
```diff
@@ -17,7 +64,6 @@ Note: `yarn global add` will default to adding packages to `~/.yarn/bin` so by d
 1.  You can add yarn's default installation location to your `$PATH` environment variable, so that your terminal knows where to look for the executables. You can do this by adding `export PATH="$(yarn global bin):$PATH"` as a line in your `~/.bashrc`. Then simply quit your terminal and reopen it (or run source ~/.bashrc) and everything should work.
 2.  Alternatively, you can change the default global location for yarn (`yarn config set prefix <filepath>`) and set `filepath` to something like `/usr/local` or `/usr` (or anything else in your `$PATH` that makes sense). This second method will require you to re-run the above global add command.
 
-Then, you must copy a `.env` file and a `keys.json` file to this folder.
```
angadgarg25: why did we remove this line?

josh-byster: Good question — I just factored it out of that portion and moved it to the top since it’s required to run both manually and with Docker. It’s moved to the top of the file.

josh-byster:
Yeah I’d like a Mac/Linux user to test this also to make sure it just works out of the box and I’m not implicitly pulling any environment variables the way I’m doing it now.

tko22:
I'm not sure if this would work with deployment using zeit now - maybe test that as well. All the current docker stuff is set up for zeit now deployment

backend/README.md
```diff
@@ -2,6 +2,53 @@
...
+```
+docker-compose up
+```
+
+And when you're done, you can run:
+```
+docker-compose down -v
+```
+
+Note that if there's a change in the `package.json` file (such as a new module added into the repository), instead of running `yarn add` to re-install, just run `docker-compose up --build`.
```
tko22: didn't know about this!

josh-byster:
> I'm not sure if this would work with deployment using zeit now - maybe test that as well. All the current docker stuff is set up for zeit now deployment

Ah, didn't see that we were using v1 deployments instead of serverless v2 functions. We can easily create a `Dockerfile-dev` in the backend that Compose pulls from instead. What are your thoughts @tko22?

josh-byster:
> The purpose of the multi stage builds is so the final image isn't big, since now has a size limit.
> 
> What I would do is I would make another Dockerfile, something like `Dockerfile.dev` and use that with docker compose.

Yes, I agree — I don't think we can use the same image built for Now since we need the dev dependencies for development. I changed the Dockerfile since I thought it was no longer used, but now that I realize it's being used to deploy BE, I will create a separate file that builds a development image.

backend/README.md
josh-byster: Yeah! Essentially I'm mounting the `./backend` and `./frontend` directories to the container for easy development, but then mounting on top of that a named volume, `fe_modules` and `be_modules` for both the frontend and backend respectively. So the process goes like this:

1. User creates Docker images for FE and BE, each with `node_modules` installed.
2. Upon starting Compose, if `fe_modules` or `be_modules` hasn't yet been created, it will grab those from the *image* and copy them to the host.
3. It will mount both the frontend and backend folders in the host to the corresponding folders in the container. This allows changes locally in code to be reflected immediately in the container.
4. It will mount over that the copied `fe_modules` and `be_modules` folders. This means that it will overwrite any local `node_modules` and replace with the copy from the image. 

This ensures uniformity on module builds among all our devs using Docker! Ta-da! :)

tko22:
> Ah, didn't see that we were using v1 deployments instead of serverless v2 functions. We can easily create a Dockerfile-dev in the backend that Compose pulls from instead, leaving the existing Dockerfile as-is. What are your thoughts @tko22?

Yep sounds good

backend/README.md
angadgarg25: ah sorry, forgot about that by the time I got down here, my bad. Looks good to go then.

PR #330, opened 2019-08-12
User API endpoints +  role change functionality for admins

### LINKED ISSUE #262: Backend for register & login workflow
Set up the backend for the registering and login workflows in the tool.

https://my-docz-project-codebrew28.hack4impact1.now.sh/register-and-login-workflow

caiyingshen:
Resolves #262

**Description:** 
This PR includes the following: 
- Internal database of users OUTSIDE of the PI auth server. The reasons for this are twofold: first, it was recommended by the PI docs that we should have our own source of truth in addition to the server in case anything goes wrong, and second, the emails that we use with our accounts may not always be representative of our names so I wanted to also include first & last name in registration so our admins can make sure they're editing the right roles. 
- Admin page functionality: the co-directors should now be able to edit user rights, as demonstrated in the gif below
ezgif-1-30ffb231ef43

### DIFF
--- a/backend/src/api/userRoles.js
+++ b/backend/src/api/userRoles.js
@@ -1,0 +1,43 @@
+const express = require('express')
+const router = express.Router()
+const { errorWrap } = require('../middleware')
+const { User } = require('../models')
+
+// get all users
+router.get(
+  '/',
+  errorWrap(async (req, res) => {
+    const workspaces = await User.find()
+    res.json({
+      code: 200,
+      result: workspaces,
+      success: true
+    })
+  })
+)
+
+// post new user
+router.post(
+  '/',
+  errorWrap(async (req, res) => {
+    const newUser = req.body
+    if (!newUser) {
+      return res.json({
+        code: 400,
+        message: 'Invalid user in request body',
+        success: false
+      })
+    }
+
+    const newBsonUser = new User({
+      user: newUser
+    })
+    await newBsonUser.save()
+
+    res.json({
+      code: 200,
+      message: `Successfully created new user ${newUser.firstName} ${newUser.lastName}`,
+      success: true
+    })
+  })
+)
--- a/backend/src/api/userRoles.js
+++ b/backend/src/api/userRoles.js
@@ -1,11 +1,12 @@
 const express = require('express')
 const router = express.Router()
-const { errorWrap } = require('../middleware')
+const { directorsOnly, errorWrap } = require('../middleware')
 const { User } = require('../models')
 
 // get all users
 router.get(
   '/',
+  [directorsOnly],
   errorWrap(async (req, res) => {
     const users = await User.find()
     res.json({
@@ -17,6 +18,7 @@ router.get(
 )
 
 // post new user
+// needed during registration, so not directorsOnly
 router.post(
   '/',
   errorWrap(async (req, res) => {
@@ -49,6 +51,7 @@ router.post(
 // update a user's role
 router.put(
   '/',
+  [directorsOnly],
   errorWrap(async (req, res) => {
     if (!req.body.email || !req.body.role) {
       return res.json({
--- a/backend/src/api/userRoles.js
+++ b/backend/src/api/userRoles.js
@@ -30,7 +30,11 @@ router.post(
     }
 
     const newBsonUser = new User({
-      user: newUser
+      firstName: newUser.firstName,
+      lastName: newUser.lastName,
+      email: newUser.email,
+      tokenId: newUser.tokenId,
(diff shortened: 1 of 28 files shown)


alan-fang:
is a user a candidate, or a member?

caiyingshen:
@alan-fang a h4i member

alan-fang:
> @alan-fang a h4i member

can you clarify what a `Pending` member is then? Based on what's been going on #311 it would just be really good to have a clear understanding of what each role here maps to.

angadgarg25:
@alan-fang Since people register themselves as users of the application, it doesn't make sense for us to give them member privileges until a director has confirmed that the person signing up is indeed a member. Therefore `PENDING` is the state a person registers in, and then they must be marked by a director in the portal before gaining `MEMBER` privileges.

alan-fang:
> @alan-fang Since people register themselves as users of the application, it doesn't make sense for us to give them member privileges until a director has confirmed that the person signing up is indeed a member. Therefore `PENDING` is the state a person registers in, and then they must be marked by a director in the portal before gaining `MEMBER` privileges.

Cool thanks for clarifying! Makes things much clearer for my work moving forwards :)

backend/src/api/userRoles.js
```diff
@@ -0,0 +1,69 @@
...
+    const newUser = req.body
+    if (!newUser) {
+      return res.json({
+        code: 400,
+        message: 'Invalid user in request body',
+        success: false
+      })
+    }
+
+    const newBsonUser = new User({
```
josh-byster: +1 for using BSON instead of JSON when referring to MongoDB storage :)

backend/src/api/userRoles.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    await User.findOneAndUpdate({ email: req.body.email }, { $set: { role: req.body.role } })
+    res.json({
+      code: 200,
+      message: 'Successfully updated user',
+      success: true
+    })
+  })
+)
```
angadgarg25: I would make all these endpoints `directors only`. Right now, I can still go in as a lead or member and change users if I manually type in `/adminRoles`.

Also lower priority since I can do this through mlab, but maybe delete functionality would be good here.

frontend/src/pages/adminRoles.js
```diff
@@ -1,64 +1,119 @@
...
+            >
+              Edit
+            </Button>
+            <Button
+              variant="info"
+              disabled={!isEditing}
+              onClick={() => this.setState({ isEditing: false })}
+            >
+              Save
+            </Button>
```
angadgarg25: Also not a huge priority, but each entry of the table has a `Submit` button, but the table as a whole also has a `Save` button.  The submit button actually completes the change, but the `Save` button simply changes the mode back to viewing. An easy fix would be renaming this to `View` or something. Eventually maybe merging the edit and view functionality, but its fine to keep it for now.

frontend/src/pages/adminRoles.js
```diff
@@ -1,64 +1,119 @@
...
+      users: []
     }
   }
 
+  componentDidMount = () => {
+    this.getUsers()
+  }
+
+  getUsers = () => {
+    getAllUsers().then(resp => {
```
tko22: could use async await syntax, but that's a nit so it's up to you;

```
getUsers = aysnc () => {
    resp = await getAllUsers()
    this.setState({
       ....
   })
}

**Comment by caiyingshen** [comment]
@angadgarg25 the server method for this endpoint requires the password of the admin so i was hesitant to add it before the roles were fully defined, but I think you're right it makes sense to add to this PR

**Comment by caiyingshen** [comment]
@angadgarg25 also forgot to mention, I don't believe the server supports delete functionality atm, so I'm hesitant to add it to our database since we might end up with data conflicts but if you don't think that's an issue for now I can add it, should be a quick fix

File: frontend/src/pages/adminRoles.js [hunk 5]
```diff
@@ -1,64 +1,201 @@
...
+    const value = event.target.value
+    const name = event.target.name
+    this.setState({ [name]: value })
+  }
+
+  handlePasswordComplete = () => {
+    const { users, selectedUser, newRole, adminPassword } = this.state
+
+    updateServerUserRole(users[selectedUser].email, newRole, adminPassword).then(resp => {
+      if (resp.status != 400) {
```
caiyingshen: to clarify, the PI server doesn't return a success value so using the status instead

caiyingshen:
new admin password modal when submitting a role change: 
Screen Shot 2019-08-20 at 11 08 12 PM

PR #332, opened 2019-08-19
Update access for different roles 

### LINKED ISSUE #314: Update login/register workflows to grant access based on roles
in the index.js/login component, users are given member access by default. We should check their role in the org and assign them their role. This will likely depend on backend implementation being complete.

caiyingshen:
Resolves #314 

**Description:** 
Allows different levels of access based on users when they sign in. 
In this PR: 
- keys.json is no longer necessary, the userId that is created by the auth server and recorded in our backend now acts as an API key between frontend and backend 
- new Pending role for new registered users. A newly registered user should be automatically directed to the pending screen (shown below) until a director grants them member/lead access 
- nav bar should update based on roles 
- members have pretty much the same rights as leads as far as I am aware, but in case I made a separate membersOnly and leadsOnly middleware in case these two diverge 

**N.B**
After this PR the login modal in the navbar that used the keys is now deprecated, you can use the following usernames/passwords (with their corresponding roles) to login: 
member@test.com
lead@test.com
director@test.com
the password for all of them is `password`. 

The tests aren't passing because of the new keys, working on fixes for those asap! 

Screen Shot 2019-08-19 at 8 33 16 PM


backend/src/middleware/membersOnly.js
```diff
@@ -0,0 +1,17 @@
+/* router middleware to only allow H4I members access
+ * must be called AFTER the auth middleware
+ */
+const membersOnly = (req, res, next) => {
+  if (!req._is_member) {
+    const msg = `The key '${req._key}' doesn't have the correct privileges. Please contact the admin.`
```
caiyingshen: this line is causing the circle ci stuff to fail, but the pre-commit hook formats it like this so if anyone knows how to temporarily turn off the hook pls lmk !!

backend/tests/utils.js
```diff
@@ -1,7 +1,7 @@
 const { Candidate, Match } = require('../src/models')
 
-const KEY = 'hjsdhfy79uu'
-const NONLEAD_KEY = 'hjsdhfy79uus'
+const KEY = '5d6846fd9289821e1277955c'
+const NONLEAD_KEY = '5d6846bf9289822ad377955b'
```
caiyingshen: userIds from the test accounts

frontend/src/pages/adminRoles.js
```diff
@@ -189,7 +192,9 @@ class AdminRoles extends React.Component {
             <Button
               color="warning"
               size="sm"
-              onClick={() => this.setState({ showPasswordModal: false })}
+              onClick={() =>
+                this.setState({ showPasswordModal: false, newRole: -1, selectedUser: -1 })
```
alan-fang: nit: what are the -1 for

backend/src/api/interview.js
```diff
@@ -1,30 +1,37 @@
...
+  '/verify_member',
   errorWrap(async (req, res) => {
     let keyVerified = false
     const key = req.query.key
+    let foundUser = null
     // removed && key.length === 11)
     if (key) {
-      keyVerified = keyData.keys.filter(currKey => currKey.key === key).length !== 0
+      foundUser = await User.findOne({ userId: key })
+      if (foundUser.email != null) {
```
alan-fang: nit: `let keyVerified = foundUser.email != null` seems cleaner

frontend/src/pages/index.js
```diff
@@ -36,18 +35,22 @@ class LoginPage extends Component {
   }
 
   handleGoogle = async e => {
-    const result = await loginGoogleUser(e.tokenId)
-    const response = await result.json()
-    if (!response.success) {
+    const response = await loginGoogleUser(e.tokenId)
+    console.log(response)
+    if (response.status != 200) {
```
alan-fang: nit: `success` not `status` (unless I'm missing some other context here)

frontend/src/pages/register.js
```diff
@@ -67,16 +64,17 @@ class RegisterPage extends Component {
   }
 
   handleGoogle = async e => {
-    const result = await loginGoogleUser(e.tokenId)
-    const resp = await result.json()
-    if (!resp.success) {
+    const resp = await loginGoogleUser(e.tokenId)
+    console.log(resp)
+    if (resp.status != 200) {
```
alan-fang: nit: `success` not `status`?

frontend/src/pages/register.js
```diff
@@ -86,14 +84,16 @@ class RegisterPage extends Component {
     if (password !== passwordVerification) {
       this.setState({ errorMessage: 'Your passwords must match.', showInvalidRequestModal: true })
     } else {
-      registerUser(email, password, permissionRolesEnum.DIRECTOR).then(resp => {
-        if (!resp.status === 400) {
+      registerUser(email, password, permissionRolesEnum.PENDING).then(resp => {
+        if (resp.status === 400) {
```
alan-fang: nit: `success` not `status`

frontend/src/pages/index.js
caiyingshen: the auth server doesn't use the success boolean apparently, so when i changed it the first time it caused some errors

frontend/src/pages/adminRoles.js
caiyingshen: mostly to show the value is supposed to be an integer and it's unassigned to a user index in the initial state

PR #333, opened 2019-08-20
upgrade nextjs to latest version. cleaner url for candidate page

tko22:
Previously, the candidate page url would be `/candidate?id=xxxxx`. With Nextjs 9, we can now do this `/candidate/xxxxx` which is a much cleaner and dynamic url. 

Other cool stuff https://nextjs.org/blog/next-9. 

from looking at the changelogs of nextjs 9, there shouldn't be anything that breaks in our recruitment portal.


frontend/src/pages/candidate/[cid].js
```diff
@@ -1,24 +1,24 @@
 import { connect } from 'react-redux'
 import React, { Component } from 'react'
 import { Container, Button, Row, Col } from 'reactstrap'
-import Router from 'next/router'
-import Head from '../components/head'
+import Router, { useRouter } from 'next/router'
```
Comment (arpanlaha) [inline frontend/src/pages/candidate/[cid].js]: where is useRouter used?

Comment (tko22) [reply frontend/src/pages/candidate/[cid].js]: not used, good catch

PR #337, opened 2019-08-20
Fix #287 add fetched list of candidates to redux, dashboard filters, tooltips for referrals

### LINKED ISSUE #287: Tooltip to show a list of people that referred/strong referred them in the dashboard view
image

A tooltip that is presented whenever a user hovers over the referral column of a specific candidate. The tooltip would show whoever referred them or if the user hovered over the strong referral column, it would show whoever strongly referred them

tooltip to use https://reactstrap.github.io/components/tooltips/

tko22:
- Resolve #287 - tooltips for people that referred them. They won't show up if there aren't any
- Add fetched List of Candidates to redux
    - `/interviews` will not fetch for candidates, unless redux store is empty. If empty, it will fetch and add to redux store
- Dashboard Filters fixed
    - Filtering by Role: Some people apply to multiple roles. Say someone applied to UI/UX and PM, if I click PM to remove people that applied to be a PM, they should still show up because they still applied for UI/UX. They should only be filtered out if I don't want to see both UI/UX and PM. 
- Search by Candidate ID in the search bar (would be useful later on, especially once we add analytics)
- Some logic is rewritten 
     
Kapture 2019-08-19 at 20 42 10

### DIFF
--- a/frontend/src/components/filterComponent.js
+++ b/frontend/src/components/filterComponent.js
@@ -146,7 +146,7 @@ class FilterComponent extends Component<Props> {
               >
                 <option selected>None</option>
                 {this.state.workspaces.map(workspace => (
-                  <option>{workspace.name}</option>
+                  <option key={workspace.name}>{workspace.name}</option>
                 ))}
               </select>
             </p>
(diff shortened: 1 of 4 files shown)


tko22:
> LOL why u logged in as me

not exactly sure why ahah

PR #338, opened 2019-08-20
Add Update Alert to dashboard

tko22:
An Alert shows up now, which would be useful for updates or notifications to users. It also saves the user's input, whenever they close the alert and would not show it again. 

To trigger a new alert (say you updated the alert description), you would then change the `AlertOff` value to something else and thus the prior cookie value would be invalid. 

image


frontend/src/pages/dashboard.js
```diff
@@ -167,185 +168,201 @@ class Dashboard extends Component {
...
+                          />
+                        </FormGroup>
+                      </Col>
+                    </Row>
+                    <Row>
+                      <Table size="m" hover className="candidate-table">
+                        <thead>
+                          <tr>
+                            <th>#</th>
+                            {selects.includes('Name') && <th>Name</th>}
```
josh-byster: Future: I know this was not within scope of your PR, so it's not a blocking change. However, can we condense all of these to a map on an array containing all the attributes (i.e. `['Name','Status','Year'].map((elem, idx) => conditionalRenderElement(elem))`
It would be great if we could also bring this out into a separate component.

frontend/src/pages/dashboard.js
```diff
@@ -167,185 +168,201 @@ class Dashboard extends Component {
...
+                                      : null}
+                                  </td>
+                                )}
 
-                              {selects.includes(selectByEnum.NUM_INTERVIEWS) && (
-                                <td> {getNumOfInterviews(candidate.interviews)}</td>
-                              )}
+                                {selects.includes('Number of Matches') && (
+                                  <td>
+                                    {candidate.facemashRankings != undefined
```
josh-byster: Nit: Can we use `&&` for conditional rendering like you did above?

frontend/src/components/dashboard/infoAlert.js
```diff
@@ -0,0 +1,29 @@
+import { Alert } from 'reactstrap'
+import cookie from 'js-cookie'
+import { useState } from 'react'
+
+const InfoAlertName = 'INFO-ALERT'
+const AlertOff = 'FALL-2019-UPDATE' // change value if you want to trigger a new alert
+
```
josh-byster: Minor: I feel like this is a good thing to have as a component that takes the expiration and inner JSX as a prop—I think it's something that could be useful to have as a wrapping component around `Alert` rather than having the infoAlert be hardcoded to use the specific "Fall 2019 Updates"

frontend/src/pages/dashboard.js
tko22: yes, we should refractor it to that!

frontend/src/components/dashboard/infoAlert.js
tko22: Yes, my original thought was that we should only use it for the dashboard, which was why it was done this way. I've updated it for us to have a general component for alerts with the cookie functionality and added a specific component for the dashboard alert. The more general component `infoAlert.js` now requires a bit more understanding on how it works for someone to actually use it though. 

Now, the point of this component (now moved to `updateAlert.js` is for us to put in some alert (say for updates) - for now, it needs to committed into the code since we don't have configuration management. But afterward, what if we want to provide updates for Spring 2020?   Then, we need to update the `AlertOff` value to be another value so the alert actually shows up.

Issue #325, opened 2019-08-08
Add index.js for components


alan-fang:
Our components folder has no `index.js` which means that to import multiple components you need to import each one directly from the component file for which it is defined in (i.e. `Nav` needs to be imported from `components/nav`. 

Add an index file and change the component imports such that multiple components are all imported from `components/`

tko22:
omg maybe we should also organzie the components folder a bit better. For example, folders like
```
- common
- candidate_page
- interview
- idk
- something else
- misc
.... 
```

PR #342, opened 2019-08-23
Resolves #335, #288 Interview Clean U + dedicated interview notes page

### LINKED ISSUE #288: Show Interview Round when displaying a summary of interviews in a candidate's page
Don't want to share a screenshot of an actual page, but I'm talking about this section in the candidates page.
image

Currently, it would show the interviewer's name and the category, if the interviewer gave them a category. Let's add the round number, somehow. One way would be group the interviews together based on their round. Or you display the round for each interview. 

Making it a table might be better. In addition, having a toast/popup with more information about the interview whenever you hover over one of the interviews would be nice.

### LINKED ISSUE #335: Interview Page
Interviews are currently shown through a Modal, which makes it much harder to link. For instance, in the candidate page, we can't link a specific interview note.

We could have something like `/interview/[interview_id]` that would render the specific interview note, link to the other notes the candidate has etc. 

OR if we did `/candidate/interviews`, we would just see the list of interviews a candidate has. 

Modals for interview should still be kept since it allows you to be on the same page while scrolling through a specific candidate's page.

tko22:
Resolves #335. Resolves #288 

Sorry for the big diff - I got bored on my flight lol

- Updates screenshot for README (just makes the status' of the candidates different)
- Interview Detail Page (#335) - triggered by /interview/[interview id]
   - Updated how interview sections are shown (reuses components in the interview page - where you actually write the notes down)
   - `Delete` Button shown if you are the interviewer that wrote those notes
       - Does not support deletion if you are director (need to wait for auth)
- Interviews in Candidate Page
   - Interview Summary 
         - interview round is displayed now
         - link to interview detail page added
    - Interview Modal    
        - View Page added per interview
- Interview Page
    - Dropdown picking category for a candidate has a default placeholder instead of placing the first value (Freshman - maybe)
    - Split page into more components (since it was a mess; it still is but ya)
- Interview Portal
  - Table shows interview round and button to view interview detail page. 
  - Edit button disabled - for now, since edit interviews isn't supported. 
- Moved common interview components to `/components/interview/` as part of #325 

### Interview detail Page
image
### Candidate Page Interview Summary
image
### Interview Modal
image

### interview portal page
image


frontend/src/components/candidates/candidateBox.js
```diff
@@ -125,33 +126,45 @@ class CandidateBox extends Component {
                 </Button>
               </Col>
             </Row>
+            <div>
```
n3a9: Use react fragment `<>`

frontend/src/pages/interview/[interviewID].js
```diff
@@ -0,0 +1,135 @@
...
+      candidateName: '',
+      verificationModalOpen: false,
+      loading: false
+    }
+  }
+
+  async componentDidMount() {
+    const query = Router.query
+    const { result, success, message } = await getInterviewByID(query.interviewID)
+    if (!success || !result) {
```
Comment (n3a9) [inline frontend/src/pages/interview/[interviewID].js]: Personally I would change order of logic here - if we have success and result then set state accordingly, else set the error message. No need for a random return either

```js
if (success && result) {
    this.setState({
       interview: result.interview,
       candidateID: result.candidateID,
       candidateName: result.candidateName
    })
} else {
    this.setState({
        errorMessage: message
    })
}

```

frontend/src/pages/interview/[interviewID].js
```diff
@@ -0,0 +1,135 @@
...
+    this.setState({
+      verificationModalOpen: false,
+      loading: false
+    })
+    Router.push('/interviewportal')
+  }
+
+  render() {
+    const { interview, candidateName, candidateID } = this.state
+    if (this.state.errorMessage !== '' || !interview) {
```
Comment (n3a9) [inline frontend/src/pages/interview/[interviewID].js]: Use conditional rendering instead of two return statements

frontend/src/pages/interview/[interviewID].js
Comment (alan-fang) [reply frontend/src/pages/interview/[interviewID].js]: I think the random return is nice because it gives the codebase better readability, do you have any other reasons why the return is bad?

frontend/src/pages/interview/[interviewID].js
Comment (alan-fang) [reply frontend/src/pages/interview/[interviewID].js]: Wouldn't that be a really ugly ternary (even if the return values were abstracted into their own components)? For readability purposes, if we keep it as is it's perfectly fine. No need to use syntactic sugar if it makes the code harder to follow

frontend/src/pages/interview/[interviewID].js
Comment (n3a9) [reply frontend/src/pages/interview/[interviewID].js]: How does a return give you 'better readability' as opposed to an else condition? The else condition allows the reader to clearly attribute the error message setting when there is some sort of failure with the result.

If the last portion of the code was setting a state for something unrelated to the initial, then the return statement makes sense. But they are part of the same logic flow so the code should be written as such. That makes it much more intuitive for someone to understand.

Comment (alan-fang) [reply frontend/src/pages/interview/[interviewID].js]: hmmm I guess you have a good point here, thanks for explaining

frontend/src/pages/interview/[interviewID].js
Comment (tko22) [reply frontend/src/pages/interview/[interviewID].js]: @n3a9 Unless the logic is completely not understandable or very hard to understand, I would say those comments are preferences, and mention it in your comment. And in those cases, especially the ones where you can enforce them using linters/formatters I would generally not mention those unless the entire repository follows the same guidelines (say the && vs null ternary)- or if I do, then I would literally prefix the comment `nit:` or something like that because it's a preference, not something that should be enforced. We all think and process differently and maybe these 2 return statements were clearer to me than you, and it would be different for someone else reading the code. 

If you do want certain preferences to be enforced (which is awesome and I think we should be more opinionated to have a standard way of writing or code all across of h4i), then you can bring it up and we can have a discussion - and if we agree to enforce it, we'll add the rule to the linter; I'd probably let you add most rules if you'd like, since it wouldn't spend much effort into those types of preferential debates.

frontend/src/pages/interview/[interviewID].js
Comment (tko22) [reply frontend/src/pages/interview/[interviewID].js]: @n3a9 Thanks for mentioning it was a personal preference! You do have a good point in that it is part of the same logic flow.

PR #345, opened 2019-08-29
Use full Node Docker image

### LINKED ISSUE #344: Use regular Node full image for development
I realized that there is really no reason to use the stripped-down Alpine image for development purposes. In addition, it was interfering with the ability to simply run `docker-compose run backend /bin/sh` and then `yarn test` because of an issue with [mongod](https://github.com/nodkz/mongodb-memory-server/issues/171). Hence, I'm proposing the change in the image used in dev (and the main frontend Dockerfile, since that's currently only being used for dev) from using Alpine to just `node:latest`.

josh-byster:
Resolves #344

### DIFF
--- a/backend/Dockerfile-dev
+++ b/backend/Dockerfile-dev
@@ -1,4 +1,4 @@
-FROM mhart/alpine-node:12.8.0
+FROM node:latest
 WORKDIR /usr/src/app
 # Copy over the module files separately so that they're cached as a build layer
 # (i.e. no rebuild necessary if these don't change) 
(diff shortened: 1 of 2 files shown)


josh-byster:
> Do you think its safe to put latest here rather than specifying a version number? I'd imagine we wouldn't run into backwards compatibility issues as the version increases (in which case its fine as is), but just wanted to check with you.

@angadgarg25 that's a good point. I suppose the better move would be to pin the version and then potentially add to the Dependabot configuration to also be checking the Dockerfiles and submit PRs when node is bumped. That's what I do for example, [here](https://github.com/josh-byster/bus-tracker/pull/63). Switching to a pinned version for this PR but will still be important since we can't run tests on alpine.

PR #347, opened 2019-08-31
Fix backend tests

### LINKED ISSUE #348: Fix backend tests
With the update in #332 where keys are now user IDs, a lot of the backend tests are breaking. We need to update the tests and fixture data so that they pass with this new framework.

caiyingshen:
Resolves #348 

<Add Screenshots!>

### DIFF
--- a/backend/src/middleware/auth.js
+++ b/backend/src/middleware/auth.js
@@ -9,6 +9,7 @@ const auth = async (req, res, next) => {
   if (key != undefined) {
     // removed && key.length === 11
     // Can add this rule if wanted
+    console.log('reached here')
 
     // check to see if URL is for register, since the key would not be recorded yet
     if (req.method === 'POST' && req.url.includes('/user/')) {
@@ -22,6 +23,7 @@ const auth = async (req, res, next) => {
         // check whether key is a director's, lead's, or member's key
         // this is used by the directorsOnly and membersOnly middleware
         if (foundUser.role === 'Director') {
+          console.log('reached here tooooooo')
           req._is_director = true
           req._is_lead = true
           req._is_member = true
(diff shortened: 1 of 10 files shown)


backend/tests/mongo_utils.js
```diff
@@ -6,30 +6,22 @@ let mongoServer
...
+  mongoServer
+    .getConnectionString()
+    .then(mongoUri => {
+      return mongoose.connect(mongoUri, {}, err => {
+        if (err) {
+          done(err)
+        }
+      })
+    })
+    .then(() => done())
```
angadgarg25: is uncommenting this what fixed it? This was the old version that I tried to rewrite with async/await with lines 19-22. Tbh idk how this even runs since `done` isn't defined anymore.

caiyingshen: Noo it fixed one of them where mongo wasn't returning the _id value of an object. When i changed the leads key to `director` instead of `lead` tho a bunch of the tests passed

backend/src/middleware/auth.js
```diff
@@ -9,6 +9,7 @@ const auth = async (req, res, next) => {
   if (key != undefined) {
     // removed && key.length === 11
     // Can add this rule if wanted
+    console.log('reached here')
 
     // check to see if URL is for register, since the key would not be recorded yet
     if (req.method === 'POST' && req.url.includes('/user/')) {
```
angadgarg25: the last failure on circle CI is probably because this line. Since `req` only has the `query` key, `req.url === undefined` so it'll throw an error like `cannot call method includes on undefined`

backend/tests/mongo_utils.js
angadgarg25: ok lets attempt to get rid of it again? if that doesn't work we can leave it, but I don't really understand how leaving this isn't causing everything to break lol

PR #349, opened 2019-09-02
Workspace TODOs

### LINKED ISSUE #305: Check user-level permissions
lots of things need to check permissions on a user-level and not role-level. wait for auth server and then do it when it is integrated into master

alan-fang:
Resolves #305 

- Added the user to the authentication layer (`auth.js`), which now allows backend to get the caller of an endpoint from the frontend.
- With that, we add an extra layer of security to cycles and workspaces, so that callers of each endpoint can only access data related to the workspace that they belong in. 
- Also added some data integrity checks to the `Users` table so that we have unique identifiers for users.
- Modified users and workspaces such that the relationship between the two entities can be one to many.
- Added backend capability to add a user to a workspace.

### DIFF
--- a/backend/src/api/cycle.js
+++ b/backend/src/api/cycle.js
@@ -25,7 +25,7 @@ router.get(
   errorWrap(async (req, res) => {
     const cycleId = req.params.cycle_id
     const cycle = await Cycle.findById(cycleId)
-    if (cycle.length == 0) {
+    if (!cycle) {
       return res.json({
         code: 404,
         message: 'Not found',
--- a/backend/src/api/cycle.js
+++ b/backend/src/api/cycle.js
@@ -50,7 +50,7 @@ router.get(
   })
 )
 
-// get all cycles belonging to a workspace that a user belongs to (either current, outdated, or both)
+// get all cycles belonging to a specified workspace. The caller must belong to that workspace.
 router.get(
   '/workspace/:workspaceName',
   [directorsOnly],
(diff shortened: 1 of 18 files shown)


backend/src/api/cycle.js
```diff
@@ -3,14 +3,12 @@ const { Cycle } = require('../models')
...
-// Get all cycles (previous and current)
-// TODO: when authentication server is integrated, only show the cycles that
-// belong to the caller of this endpoint.
+// Get all cycles belonging to the same workspace that a User is in (previous and current)
 router.get(
   '/',
   [directorsOnly],
   errorWrap(async (req, res) => {
-    const cycles = await Cycle.find()
+    const cycles = await Cycle.find({ workspaceName: { $in: req._user.workspaceIds } })
```
caiyingshen: is there a diff between `workspaceName` and `workspaceIds`? if not, I think it would make more sense if we only used one of them to prevent confusion but this is more a preference/not a blocker

alan-fang: There's more stuff to be done so I can address it in a follow-up

backend/src/api/cycle.js
```diff
@@ -36,28 +50,25 @@ router.get(
   })
 )
 
-// get all cycles belonging to a workspace (either current, outdated, or both)
-// TODO: when authentication server is integrated, only show the cycles
-// to the workspace's owner. Otherwise, 403.
+// get all cycles belonging to the workspace that a user belongs to (either current, outdated, or both)
 router.get(
-  '/workspace/:workspaceName',
+  '/workspace',
```
chyku: note to self: the way we get workspaces in the frontend is going to have to be changed to getting ids from user

alan-fang: I changed the rendering issues on the frontend already

alan-fang:
Re-requesting review because the changes since last approval are nontrivial

PR #350, opened 2019-09-02
Update README.md

alan-fang:
Added a blurb about new login information so that `keys.json` doesn't confuse anyone else doing dev work in the future.

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -34,7 +34,6 @@ However, `keys.json` is no longer used as part of the login. We use the followin
 
   * member@test.com
   * lead@test.com
-  * director@test.com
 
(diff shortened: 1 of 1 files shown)


backend/README.md
```diff
@@ -30,6 +30,14 @@ In `keys.json`:
 }
 ```
 
+However, `keys.json` is no longer used as part of the login. We use the following accounts as test logins:
+
+  * member@test.com
+  * lead@test.com
+  * director@test.com
```
caiyingshen: I think it's safer to not include `director@test.com` since directors have rights have access to our auth server and change it so if someone runs this and uses that login they could potentially make changes to our orgs users.

PR #369, opened 2020-07-06
Dockerize Mongo

### LINKED ISSUE #367: Dockerize Database Layer
**Summary**
Making the database run as a container in Docker which can be quickly spun up or destroyed locally, allowing us to quickly develop in isolation.

**Problem / Current Functionality**

Currently, we have this hybrid approach where some things are containerized in Docker and some parts (like the database) are shared among all developers in the cloud. This means that everything is connecting to this centralized instance (currently some MLab URL put in our `env` file). As a result, if developer A is doing something in development, it affects what happens to person B in their development environment. That can be counterproductive to development, and doesn't allow for the flexibility of being able to modify any and everything at-will.

For a true Docker experience, every piece of development should live & be isolated in a local container. Currently this is only the case for our frontend and backend.

**Solution**

[LAH Docker Compose file](https://github.com/hack4impact-uiuc/life-after-hate/blob/master/docker-compose.yml) can be a reference point for this.

Let's create a Docker container that holds inside it the Mongo database. This will be defined most likely as a service in the `docker-compose.yml` file. We'll start by using one of the existing [MongoDB images](https://hub.docker.com/_/mongo). When a container is created from this Mongo image, it'll automatically start listening on a port (ex. `27017`).  We want to [publish](https://docs.docker.com/config/containers/container-networking/#published-ports) that to the outside world so that it can be accessible outside so we can access it from our host computer. 
When this step is set up correctly, one should be able to run `docker-compose up` and go into an application like MongoDB Compass and successfully connect to `mongodb://localhost:27017/`. 

Next, we'll need to connect this to the application. To do this, we'll make sure our backend `depends_on` the database container, so that it first will launch the database before the backend starts up and tries to connect (or else it would be possible for the backend to have started before the database and fail to connect since the database hasn't started). We can reference the database by using it's container name, i.e. `mongodb://db:27017/` since it's within the same [network](https://docs.docker.com/compose/networking/) so it'll look up the container's IP automatically.

Lastly, we want to make sure we're mounting a volume into the container, so that the data will be stored on our host and is persistent across sessions. This way, when you run `docker-compose down`, the database data is still stored safely on the host even though the container went down. Next time `docker-compose up` is run, it'll mount that directory containing the data into the new container, and then that new container will read & write to that data.
We will be able to destroy the volume using `docker-compose down -v`. 

**Acceptance Criteria / User Stories**
- [x] As a platform developer, I should be able to run `docker-compose up` and have it spin up a database which is used by the platform for all Mongo requests.
- [ ] As a platform developer, there's clear documentation on how to spin up & down this database (or an explanation on how it works to someone not familiar with Docker.

alicesf2:
Resolves #367 

- Added the db service to the `docker-compose.yml` file
- Removed all commands from ./recreate_db.sh except for `dotenv yarn populatedb`
- Removed unused backend .env variables

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -7,35 +7,51 @@ In order to have all the secret keys set up properly, you must create a `.env` f
 Example setup:
 
 In `.env`:
+
 ```
-MONGO_UURL=mongodb://mongouser:mykey@somedb.mlab.com:port/db
-MONGO_DURL=mongodb://mongouser:mykey@somedb.mlab.com:port/db
-MONGO_URL=mongodb://mongouser:mykey@somedb.mlab.com:port/db
+MONGO_URL=mongodb://db:27017/endpoint
 SCHEDULER_URL=https://scheduler-URI.amazonaws.com/endpoint
 SCHEDULER_API_KEY=abcd1234
 ```
 
 We use the following accounts as test logins:
 
-  * member@test.com
-  * lead@test.com
+- member@test.com
+- lead@test.com
 
 The password for all of them is `password`.
 
 ### Docker
 
-To run with [Docker](https://www.docker.com/) (and start the frontend and backend all at once with one command):
+To run with [Docker](https://www.docker.com/) (and start the frontend, backend, and database all at once with one command):
+
 ```
 docker-compose up
 ```
 
 And when you're done, you can run:
+
+```
+docker-compose down
+```
+
+to stop and remove all docker containers, or
+
 ```
 docker-compose down -v
 ```
 
-Note that if there's a change in the `package.json` file (such as a new module added into the repository), instead of running `yarn add` to re-install, just run `docker-compose up --build`. 
+to remove all docker volumes, which also deletes all the data stored in the database.
 
+If you previously ran `docker-compose down -v` or just cloned the repo, you'll also need to run:
+
+```
+./recreate_db.sh
+```
+
+which repopulates the database with mock data.
+
+Note that if there's a change in the `package.json` file (such as a new module added into the repository), instead of running `yarn add` to re-install, just run `docker-compose up --build`.
 
 ### Manual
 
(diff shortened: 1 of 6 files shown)


josh-byster:
Looks great, Alice! 🚀 

Small thing: to allow us to populate the DB and have the full 
"Docker experience", let's do as much as we can inside the Docker container itself. Hence it would make sense to spin up a container and seed the database from inside that container. To do this we can change the `populatedb` script in the package JSON file to `docker-compose run backend node src/utils/populate.js`, which will spin up a container, run the same command inside the container, and then remove the container.

backend/README.md
```diff
@@ -7,35 +7,51 @@ In order to have all the secret keys set up properly, you must create a `.env` f
 Example setup:
 
 In `.env`:
+
 ```
-MONGO_UURL=mongodb://mongouser:mykey@somedb.mlab.com:port/db
-MONGO_DURL=mongodb://mongouser:mykey@somedb.mlab.com:port/db
-MONGO_URL=mongodb://mongouser:mykey@somedb.mlab.com:port/db
+MONGO_URL=mongodb://db:27017/endpoint
```
josh-byster, line 12: Nit: maybe renaming `endpoint` to something else to not confuse, as what comes after the `/` will be the [default database name](https://docs.mongodb.com/manual/reference/connection-string/) rather than a API-like request URL.

backend/README.md
```diff
@@ -7,35 +7,51 @@ In order to have all the secret keys set up properly, you must create a `.env` f
...
+
 ```
 docker-compose down -v
 ```
 
-Note that if there's a change in the `package.json` file (such as a new module added into the repository), instead of running `yarn add` to re-install, just run `docker-compose up --build`. 
+to remove all docker volumes, which also deletes all the data stored in the database.
+
+If you previously ran `docker-compose down -v` or just cloned the repo, you'll also need to run:
+
```
josh-byster, line 47: Love the added clarity here! :)

backend/README.md
alicesf2: ohh i see, it shouldn't be the actual mongo_url tho cuz that's a security issue? so i should do smth like /<default-db-name>?

PR #370, opened 2020-07-08
Continuous integration

### LINKED ISSUE #368: Continuous Deployment via Vercel
**Summary**
We want a way to be able to quickly deploy the entire application stack with one command, allowing us to be fully running and be able to plug in the environment variables (ex. database URI) on the Vercel website.

**Problem / Current Functionality**
Currently, we are able to deploy parts of our application separately with Vercel/Now, but it's not super easy to be able to do so. We also currently use [version 1](https://github.com/hack4impact-uiuc/h4i-recruitment/blob/master/backend/now.json) for deployments, which is deprecated and will be removed soon. So we want to upgrade everything to use the latest Vercel serverless functions and deploy both simultaneously using a `now.json` file in the root of the repo.

**Solution**

The LAH `now.json` [file](https://github.com/hack4impact-uiuc/life-after-hate/blob/master/now.json) may be helpful as reference.
Using a local tool like `vercel dev` may be extremely helpful when getting this set up. We need to have the routing all set up with a `vercel.json` file, so that requests going to, say, `https://recruitment-portal.vercel.sh/api/` get forwarded to the backend serverless function, while requests that don't match that format get forwarded to the frontend, where the static assets will be served from edge caches located globally.

We'll also want to create instructions on how to create, say, a MongoDB Atlas instance and deploy that as well. The Vercel deployment should be located in the same AWS region as the Atlas deployment for low latency. For example, a database deployed in N. Virginia should have `"regions": ["iad1"]` in the Vercel configuration file to deploy in the same region.

**Acceptance Criteria / User Stories**
- [ ] As a superadmin, I should be able to deploy the application to my own account using `vercel` in the CLI and set up the environment variables in the server.

coconut750750:
Resolves #368


josh-byster:
Awesome, thanks @coconut750750! Any idea why the styling is off in the deployed link?

backend/src/app.js
```diff
@@ -36,7 +36,7 @@ app.use(morgan('dev'))
 // verifies user
 app.use(auth)
 
-app.use('/', routes)
+app.use('/api/', routes)
```
kenetec, line 39: Won't this make the URL to access the backend be /api/api?

josh-byster: Great question—it shouldn't, specifically because right now they're of the form `http://localhost:8080/interviews/` for example if running locally. This worked great when we would have, for example, `https://my_backend_server.heroku.com/interviews` running on a different domain from the frontend server. That way we could have `https://my_backend_server.heroku.com/interviews` for the API and `https://my_frontend_server.heroku.com/interviews` for the frontend and there would be no conflict. But now, since we'll be hosting them all on one Vercel domain, we need to route the two separately, so have everything that starts with `/api` be routed to our Express functions, while everything else be routed to the frontend.

coconut750750:
Hm, I don't know why the styling is off. It seems like it's failing to load the `style.css` in the `static/` directory. Any ideas why or how to fix this?

josh-byster:
Yeah, you're absolutely right. After some investigation, it looks like since we switched to using the root project directory instead of running `next` in the context of the `src` directory, the `static` folder wasn't getting recognized. One of the exceptions to using `src` with this change is that the `static` or `public` folder must be [outside](https://nextjs.org/docs/advanced-features/src-directory#caveats) the `src/` folder, so I made that edit. 

API tests are all failing since we moved to using `/api/...` for our routes. I think it would be great if we can figure out a way to do this without having to change every single test written. There's gotta be a way in supertest to specify a base API URI for what we're doing, but some preliminary searching turned up nothing. Would you mind digging into this a bit @coconut750750? Worst case we just change everything from `/user` to `/api/user` for example, but it would be great if we could keep this abstracted away.

coconut750750:
@josh-byster I dont think there is a way. On SO, there are some solutions for remote APIs: https://stackoverflow.com/questions/13754105/reuse-supertest-tests-on-a-remote-url. 

I guess we should just change all of them to hit `/api/`

josh-byster:
While unfortunate, that makes sense. Thanks for looking into it and making the fixes, @coconut750750!

